PR #1052, opened 2015-07-22
Probably a non-issue but it remove a warning :)

ker2x:
cqwindow.c:562:2: warning: ‘timeRes’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  return timeRes;
  ^
cqwindow.c:562:2: warning: Undefined or garbage value returned to caller
        return timeRes;
        ^~~~~~~~~~~~~~

If none of the branch are taken then timeRes contain garbage value (non initialized).
I propose to initialize it to NULL and add an Assert, just to be sure :)


usmanm:
Merged. Thanks man--we appreciate you contributing! :)

ker2x:
I'm planning to use this software in production someday so it's a win-win situation :)

PR #1054, opened 2015-07-22
Custom data types improvements and bug fixes

usmanm:
- Now supports in/out functions so can be used with `COPY`
- Fixed a few segfaults where the datatype was `NULL`
- Adding an element to a `NULL` value automatically initializes a default empty data structure
- Hashing fix for `text` type when inserted from `psql` which was previously failing because polymorphic type coercion fails with `unknown` types
- Added a print function for all types
- A few `nargs` bugs in `pg_proc` catalog

### DIFF
--- a/src/backend/pipeline/bloom.c
+++ b/src/backend/pipeline/bloom.c
@@ -43,6 +43,8 @@ BloomFilterCreateWithMAndK(uint32_t m, uint16_t k)
 	bf->k = k;
 	bf->blen = blen;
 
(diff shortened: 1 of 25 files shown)


derekjn:
Looks good!

usmanm:
The slight downside is that for t-digest types, it now fixes the size of the whatever the max "size" of the digest can be. In our case, the default is 200. What this means is that for t-digests where < 200 values have been added, we now use more space. Figured < 200 is probably not a common case.

PR #1062, opened 2015-07-24
Matrel vac improvements

derekjn:
Fixes some matrel bloating issues. There were a few things going on here:

* Long running-xacts block autovac garbage collection, so long-running stream insertions will now periodically commit to prevent this from happening
* Table stats weren't being reported often enough after `sync_combine`, which prevented that autovac from realizing that a matrel was due for a vacuum
* Since we don't call `heap_delete` when deleting an expired SW tuple, a tuple's TOAST rows weren't being marked for deletion

The default naptime for the autovac also isn't ideal for update-heavy matrels, so I globally lowered it from `60s` to `15s`. It's not a table-level parameter.

### DIFF
--- a/src/backend/pipeline/cont_combiner.c
+++ b/src/backend/pipeline/cont_combiner.c
@@ -995,6 +995,7 @@ next:
 		}
 
 		CommitTransactionCommand();
+		pgstat_report_stat(false);
 
 		if (num_processed)
 			last_processed = GetCurrentTimestamp();
--- a/src/backend/pipeline/streamReceiver.c
+++ b/src/backend/pipeline/streamReceiver.c
@@ -9,6 +9,9 @@
  * IDENTIFICATION
  *	  src/backend/pipeline/streamReceiver.c
  */
+#include "postgres.h"
+
+#include "access/xact.h"
 #include "pipeline/stream.h"
 #include "pipeline/streamReceiver.h"
 #include "pipeline/tuplebuf.h"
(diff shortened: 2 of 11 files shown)


src/backend/pipeline/cont_combiner.c
```diff
@@ -995,6 +995,7 @@ ContinuousQueryCombinerMain(void)
 		}
 
 		CommitTransactionCommand();
+		pgstat_report_stat(true);
```
usmanm: Lets not force this. This function will automatically force once every `PGSTAT_STAT_INTERVAL`.

src/backend/pipeline/streamReceiver.c
```diff
@@ -39,6 +44,21 @@ stream_receive(TupleTableSlot *slot, DestReceiver *self)
...
 	}
 
+	/*
+	 * Writing to streams from a long-running process is presumably a common case,
+	 * so we periodically commit to avoid a long-running transaction. Long-running xacts
+	 * blocks the autovac from freeing up space, because tuples that have long since been
+	 * deleted would technically still be visible from the point of view of a long-running
+	 * xact.
+	 */
+	if (TimestampDifferenceExceeds(GetCurrentTimestamp(), stream->lastcommit, STREAM_INSERT_COMMIT_INTERVAL))
```
usmanm: We should probably make this into a GUC parameter, with the option of disabling it by setting interval to -1. In cases where there are things other than stream insertions also in the transaction, this behavior might not be ideal.

usmanm:
Nice dude!

src/backend/pipeline/sw_vacuum.c
```diff
@@ -234,6 +234,13 @@ ShouldVacuumSWTuple(SWVacuumContext *context, HeapTupleData *tuple)
 	if (!context)
 		return false;
 
+	/*
+	 * The combiner could have updated this tuple before actually committing/aborting,
+	 * so we ignore it in that case.
+	 */
+	if (tuple->t_data->t_infomask & (HEAP_XMAX_COMMITTED | HEAP_XMAX_INVALID))
```
usmanm: Should this be `HEAP_XMIN_COMMITTED | HEAP_XMIN_INVALID`? 

`xmax` is only set when the tuple is deleted, or has been replaced by an updated tuple. `xmin` is the xid of the transaction that created the tuple.

http://www.postgresql.org/docs/9.1/static/ddl-system-columns.html

derekjn:
Can you give this one more once over? I made a few changes and there are still some potential issues, but this fixes most of what we were seeing and should be fine until #1068 puts all of this to rest.

usmanm:
Lgtm!

src/backend/utils/misc/guc.c
```diff
@@ -2668,6 +2669,17 @@ static struct config_int ConfigureNamesInt[] =
 		NULL, NULL, NULL
 	},
 
+	{
+		{"stream_insertion_commit_interval", PGC_BACKEND, QUERY_TUNING_OTHER,
+		 gettext_noop("Sets the default amount of time to periodically commit from a stream insertion process."),
+		 gettext_noop("A lower number will minimize the amount of time the autovacuumer can be blocked from freeing unused space by long-running stream insertion processes."),
```
usmanm: Maybe we should also add that a lower number will make sure newly created CVs are seen by the insert process faster, otherwise if a CV is created in the middle of a long running INSERT transaction, it will not see any data.

PR #1070, opened 2015-07-26
Create dependency between pipeline_combine entry and pg_aggregate func

usmanm:


derekjn:
Looks good. We need to think through migration automation, because this (and upcoming stuff) is going to break existing installations.

Issue #1057, opened 2015-07-23
pipeline shell (psql) default port doesn't works when postgresql is installed


ker2x:
When postgresql client/common is installed on the system, or more specifically the libpq is installed system-wide then pipeline try to connect to port 5432 or use the default postgresql .s.PGSQL.5432 unix socket.

The thing is : the default port/socket/everything are hardcoded in libpq.

It also try to use my ~/.pgpass and all stuff from http://www.postgresql.org/docs/9.4/static/libpq-envars.html

I guess a choice need to be done somehow : 
- use all defaults from postgresql
- or change everything (including default env var, default file, etc)

pipeline --help say : 
>  -p, --port=PORT          database server port (default: "6543")

but running pipeline : 
>   pipeline: could not connect to server: Aucun fichier ou dossier de ce type
>   	Is the server running locally and accepting
>   	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?

and pipeline -h 127.0.0.1 :
>   pipeline: could not connect to server: Connexion refusée
>   	Is the server running on host "127.0.0.1" and accepting 
>   	TCP/IP connections on port 5432?

pipeline -h127.0.0.1 -p6543 -d pipeline :
> pipeline (9.4.4)
> Type "help" for help.
>
> pipeline=#

derekjn:
The problem here is that `pipeline` symlinks to `/usr/bin/psql` upon installation. I think the correct thing to do here is just to link `/usr/bin/pipeline` to `<install directory>/bin/psql` instead of the system-wide binary.

ker2x:
hummm nope. not in my case. i used --prefix to install it to a development directory and it used the pipeline version of psql : 

> ker2x@lily:~/build/bin$ pipeline --help
> pipeline is the PipelineDB interactive terminal.
> 
> Usage:
> [...]
> Connection options:
>  -h, --host=HOSTNAME      database server host or socket directory (default: "local socket")
>  -p, --port=PORT          database server port (default: "6543")
>  -U, --username=USERNAME  database user name (default: "ker2x")
>  -w, --no-password        never prompt for password
>  -W, --password           force password prompt (should happen automatically)
> 
> For more information, type "\?" (for internal commands) or "\help" (for SQL
commands) from within psql, or consult the psql section in the PostgreSQL
documentation.
> 
> Report bugs to <eng@pipelinedb.com>.

derekjn:
[...]

PR #1071, opened 2015-07-27
override libpq default port

ker2x:
see #1057


derekjn:
Thanks @ker2x ! Do you mind signing our CLA so we can accept any pull requests you send our way?

ker2x:
sure, no problem. i'll see that tomorrow :)

ker2x:
doh!?

ker2x:
i can see how my patch break stuff.

It act as we passed -p option so it also override environment var and psql rc file.

This patch sux, sorry :)

PR #1073, opened 2015-07-27
Remove empty_wait configuration parameter

usmanm:
Don't think it's that useful to tune it, we just use `max_wait` as the sleep time now. Also increased the default from 5ms to 25ms.


derekjn:
Can you update `pipelinedb.conf.sample`? Otherwise looks good!

PR #1077, opened 2015-07-28
Randomize ID generation of continuous views

usmanm:
It's not random in the range `[1, MAX_CQS)`. Makes some sort of effort to keep the ID range compact, but retain enough slack.

- Also removes the empty_sleep config param
- Fixes a few tests


usmanm:
Hold off on it. I need to add some logic that tries to uniformly distribute queries to all combiners.

usmanm:
Good to go.

derekjn:
Looks good!

PR #1079, opened 2015-07-29
Continuous query processes should be nice

usmanm:


src/backend/pipeline/cont_scheduler.c
```diff
@@ -47,6 +48,7 @@
 #define INIT_PROC_TABLE_SZ 4
 #define TOTAL_SLOTS (continuous_query_num_workers + continuous_query_num_combiners)
 #define MIN_WAIT_TERMINATE_MS 250
+#define CONT_PROC_PRIORITY 0.75
```
derekjn: Can we make this a guc param?

derekjn:
That was easy, looks good

src/backend/pipeline/cont_scheduler.c
usmanm: Good call!

PR #1082, opened 2015-08-04
1078 adhoc client

loadzero:
Finished work on the ncurses adhoc query client for the time being.

The binary is called ```padhoc```

Run it like this

```
./bin/generate-adhoc | padhoc
```

The keys are

```
UP         ScreenScrollUp
DOWN       ScreenScrollDown
LEFT       ScreenScrollLeft
RIGHT      ScreenScrollRight
NPAGE      ScreenPageDown
HOME       ScreenHome
END        ScreenEnd
PPAGE      ScreenPageUp
TAB        ScreenNextCol
SHIFT+TAB  ScreenPrevCol
p          ScreenTogglePause
```
ctrl-c will quit.

TODO: 

* hook up to the backend
* optimise the rowmap
* add tests
* unicode support


src/bin/psql/adhoc_compat.h
```diff
@@ -0,0 +1,31 @@
...
+
+void
+ExceptionalCondition(const char *conditionName,
+					 const char *errorType,
+					 const char *fileName,
+					 int lineNumber);
+
+void elog_start(const char *filename, int lineno, const char *funcname);
+void elog_finish(int elevel, const char *fmt,...);
+
```
derekjn: Why are these stubs necessary?

src/bin/psql/adhoc_compat.h
```diff
@@ -0,0 +1,31 @@
...
+typedef struct FlexString {
```
derekjn: Is there a reason you're not reusing `PQExpBufferData` in `libpq/pqxpbuffer.h`? It's an identical `struct`.

src/bin/psql/adhoc_compat.c
```diff
@@ -0,0 +1,78 @@
...
+{
+	assert(0);
+	abort();
+}
+
+void init_flex(FlexString* f)
+{
+	memset(f, 0, sizeof(FlexString));
+
+	f->buf = malloc(1025);
```
derekjn: Probably should be using `pg_malloc` like `psql` does.

src/bin/psql/model.c
```diff
@@ -0,0 +1,174 @@
...
+}
+
+static size_t FindInRow(Row *r, const char* s)
+{
+	size_t i = 0;
+
+	for (i = 0; i < RowSize(r); ++i)
+	{
+		if (strcmp(RowFieldValue(r,i), s) == 0)
+		{
```
derekjn: Per the style guide, brackets aren't necessary for single statements under control structures:

```
if (strcmp(RowFieldValue(r,i), s) == 0)
    break;
```

http://www.postgresql.org/docs/current/static/source.html

src/bin/psql/model.c
```diff
@@ -0,0 +1,174 @@
...
+	RowCleanup(r);
+}
+
+static size_t FindInRow(Row *r, const char* s)
+{
+	size_t i = 0;
+
+	for (i = 0; i < RowSize(r); ++i)
+	{
+		if (strcmp(RowFieldValue(r,i), s) == 0)
```
derekjn: `pg_strcasecmp`

src/bin/psql/model.c
```diff
@@ -0,0 +1,174 @@
...
+			break;
+		}
+	}
+
+	return i;
+}
+
+void ModelSetKey(Model *m, Row* r)
+{
+	// key names must exist in the header.
```
derekjn: C++ style comments aren't used anywhere, as per the style guide.

src/bin/psql/model.c
```diff
@@ -0,0 +1,174 @@
...
+
+	RowKeyReset();
+
+	for (i = 0; i < RowSize(r); ++i)
+	{
+		size_t ki = FindInRow(&m->header, RowFieldValue(r, i));
+
+		if (ki == RowSize(&m->header))
+		{
+			die("could not find val");
```
derekjn: Try to be as descriptive as possible with any error messages. This seems a bit cryptic.

src/bin/psql/model.c
```diff
@@ -0,0 +1,174 @@
...
+
+	printf("nfield %zu\n", m->nfields);
+
+	for (i = 0; i < m->nfields; ++i)
+	{
+		printf("field %zu %zu\n", i, m->maxlens[i]);
+	}
+}
+
+size_t spaces(const char* s);
```
derekjn: `const char* s` -> `const char *s`

src/bin/psql/model.h
```diff
@@ -0,0 +1,32 @@
...
+{
+	RowMap *rowmap;
+	size_t* maxlens;
+	size_t nfields;
+
+	Row header;
+
+} Model;
+
+#define MIN(a, b)  (((a) < (b)) ? (a) : (b))
```
derekjn: These are already available as `Min` and `Max`, which are defined in `c.h`, which is imported via `postgres_fe` in `psql`.

src/bin/psql/rowmap.c
```diff
@@ -0,0 +1,357 @@
...
+		return 1;
+	}
+
+	if (RowSize(r2) != g_row_key_n)
+	{
+		RowDump(r1);
+		RowDump(r2);
+	}
+
+	assert(RowSize(r2) == g_row_key_n);
```
derekjn: Use `Assert`.

src/bin/psql/adhoc.c
```diff
@@ -0,0 +1,135 @@
...
+		ScreenDestroy(screen);
+	}
+
+	RowStreamDestroy(stream);
+	ModelDestroy(model);
+
+	return 0;
+}
+
+void row_callback(void* ctx, int type, Row* row)
```
derekjn: As per the style guide, function return types go on their own line:

```
void
row_callback(void* ctx, int type, Row* row)
```

derekjn:
Apart from the mostly stylistic issues, the only other general feedback I have at this point is that there are almost no comments in this code. Generally, nearly every function should have a header with a brief description about what it does, unless it's something blatantly obvious like a very short private `static` function.

If you open pretty much any file in the project you'll see the same general density of comments, so that's a good baseline to go off of.

bin/generate-adhoc
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env ruby
```
usmanm: If possible, can we port this to Python? Not that it matters too much, but adding more languages to the code base is always a slippery slope.

src/bin/psql/adhoc.c
```diff
@@ -0,0 +1,135 @@
...
+	Model* model;
+	Screen* screen;
+} App;
+
+volatile int keep_running = 1;
+void sighandle(int);
+
+void sighandle(int x)
+{
+	keep_running = 0;
```
usmanm: `keep_running = false`

src/bin/psql/adhoc.c
```diff
@@ -0,0 +1,135 @@
...
+}
+
+void row_callback(void* ctx, int type, Row* row);
```
usmanm: Is there a reason `ctx` isn't an `App *` type?

src/bin/psql/adhoc.c
```diff
@@ -0,0 +1,135 @@
...
+	switch (type) {
+
+		case 'k':
+			ModelSetKey(app->model, row);
+			break;
+		case 'h':
+			ModelSetHeader(app->model, row);
+			if (app->screen) ScreenUpdate(app->screen);
+			break;
+		case 'i': case 'u':
```
usmanm: Cases should be in different lines. Also we're calling `ModelAddRow` for both insert and update events? Is that intended?

src/bin/psql/adhoc_compat.c
```diff
@@ -0,0 +1,78 @@
...
+{
+}
+
+void elog_finish(int elevel, const char *fmt,...)
+{
+	assert(0);
+	abort();
+}
+
+void init_flex(FlexString* f)
```
usmanm: This seems very similar to `StringInfo`. Do we not have access to that in `psql`?

src/bin/psql/model.c
```diff
@@ -0,0 +1,174 @@
...
+
+	return m;
+}
+
+void ModelDestroy(Model *m)
+{
+	RowMapDestroy(m->rowmap);
+	RowCleanup(&m->header);
+
+	free(m->maxlens);
```
usmanm: `pfree` or whatever it's equivalent in `psql` is.

src/bin/psql/model.c
```diff
@@ -0,0 +1,174 @@
...
+
+void ModelAddRow(Model *m, Row* r)
+{
+	ModelUpdateLens(m, r);
+	RowMapUpdate(m->rowmap, r);
+}
+
+void ModelDeleteRow(Model *m, Row* r)
+{
+	RowMapErase(m->rowmap, r);
```
usmanm: Do we not need to update the lens here? In case a row with the max length for a field is being removed?

src/bin/psql/model.c
```diff
@@ -0,0 +1,174 @@
...
+	}
+}
+
+size_t spaces(const char* s);
+
+size_t spaces(const char* s)
+{
+	size_t cnt = 0;
+
+	while (*s)
```
usmanm: Maybe change this to `while (*s != '\0')` for clarity that we're finding the string terminator.

src/bin/psql/model.c
```diff
@@ -0,0 +1,174 @@
...
+			cnt++;
+		}
+
+		s++;
+	}
+
+	return cnt;
+}
+
+Row make_row(const char* s);
```
usmanm: Not sure if we need this declaration.

src/bin/psql/model.c
```diff
@@ -0,0 +1,174 @@
...
+
+Row make_row(const char* s)
```
usmanm: Can mark this as `static`?

src/bin/psql/model.h
```diff
@@ -0,0 +1,32 @@
+#ifndef MODEL_H_4D955674
```
usmanm: What's this random hex at the end?

src/bin/psql/model.h
```diff
@@ -0,0 +1,32 @@
...
+	size_t nfields;
+
+	Row header;
+
+} Model;
+
+#define MIN(a, b)  (((a) < (b)) ? (a) : (b))
+#define MAX(a, b)  (((a) > (b)) ? (a) : (b))
+
+Model* ModelInit(void);
```
usmanm: Add an `extern` before this declarations. Postgres convention--don't think it matters.

src/bin/psql/rowmap.c
```diff
@@ -0,0 +1,357 @@
...
+}
+
+static inline size_t key_length(Row *r)
+{
+	size_t n = 0;
+	size_t i = 0;
+
+	for (i = 0; i < g_row_key_n; ++i)
+	{
+		n += r->fields[i].n;
```
usmanm: `n += r->fields[i].n + 1;`

src/bin/psql/rowmap.h
```diff
@@ -0,0 +1,62 @@
...
+void RowKeyAdd(size_t i);
+
+size_t RowFieldLength(Row* r, size_t i);
+char* RowFieldValue(Row* r, size_t i);
+Field* RowGetField(Row* r, size_t i);
+
+Row RowGetKey(Row* r);
+
+typedef struct RowMap {
+
```
usmanm: No need for blank lines at the beginning and end of a `struct` definition.

usmanm:
Nice dude, this looks solid! Didn't know about `ncurses` but it looks pretty nifty. Similar feedback on adding some more comments to understand how all these pieces fit together.

src/bin/psql/model.h
loadzero: Its a uuid of sorts - I have been bitten by this before, where two headers have the same short name (and same ifdef guard). So by default, I set up my editor to add a unique guard.

```
#include "a/foo.h"
#include "b/foo.h"
```

src/bin/psql/adhoc_compat.h
loadzero: This was a leftover from pulling in the backend rbtree. Not needed anymore, I will kill them off.

src/bin/psql/adhoc_compat.h
loadzero: Nope - good call, changed.

src/bin/psql/adhoc_compat.c
loadzero: yep, changed.

src/bin/psql/model.c
loadzero: Good point - I need to use pg_strcasecmp to get postgres string sort ordering. I will revisit this when I tackle ordering in the backend.

src/bin/psql/model.c
loadzero: Fixed

src/bin/psql/rowmap.c
loadzero: I have fixed this and added an adhoc client specific assert, because we need to clean up ncurses before displaying the assert message. (see TermAssert in adhoc_util.h)

src/bin/psql/adhoc.c
loadzero: Fixed

src/bin/psql/adhoc.c
loadzero: This is done on purpose - the callback is defined generically in RowStream.

Another way to do this would be to cast the function ptr, but I think this reads cleaner.

src/bin/psql/adhoc.c
loadzero: It was intended, but I have now refactored it to read cleaner.

src/bin/psql/adhoc_compat.c
loadzero: No, but we do have access to PQExpBuffer, I have fixed the code to use that instead.

src/bin/psql/model.c
loadzero: fixed

src/bin/psql/model.c
loadzero: No, this is an intended effect, to reduce screen jumpiness.

We could make this a configurable option, or add a key to 'shrink' columns to fit.

src/bin/psql/model.c
loadzero: Fixed

src/bin/psql/model.c
loadzero: Fixed - made a bunch of stuff static

src/bin/psql/model.h
loadzero: fixed

loadzero: Fixed

src/bin/psql/rowmap.h
loadzero: Fixed

Issue #1046, opened 2015-07-16
Add all dependencies to continuous views


usmanm:
- columns of relations being joined against

derekjn:
Functions used in continuous view overlays.

PR #1088, opened 2015-08-09
#1046: add all necessary dependencies for CVs

derekjn:
Well that was easy. Definitely had some dependencies bugs that this uncovered.


usmanm:
Looks good!

PR #1089, opened 2015-08-09
1081 stats

derekjn:
Adds global stats that live for the duration of the database. Each database's stats file is timestamped when it is first seen, which should be good enough for uniqueness.

This also eliminates the `libcurl` dependency.


src/backend/pipeline/update.c
```diff
@@ -31,114 +30,204 @@
...
- * 0) Operating system
- * 1) Number of continuous views
- * 2) Aggregate input/output rows/bytes
- */
-void
-UpdateCheck(HTAB *all_dbs, bool startup)
+static int
+post(char *payload)
+{
+  struct hostent *server;
```
usmanm: Fix indentation.

usmanm:
Looks good! +1 on removing `libcurl`.

src/backend/pipeline/update.c
```diff
@@ -31,114 +30,204 @@
 #include "utils/rel.h"
 #include "utils/syscache.h"
 
+#define ANON_ENDPOINT		"localhost"
```
usmanm: Might want to change this to: `"http://anonymous.pipelinedb.com/check"`?

usmanm:
Maybe we should pass `-c anonymous_update_checks=false` to `pg_regress`, Python tests and `make run`.

src/backend/pipeline/update.c
derekjn: Haha, that is probably a good idea.

PR #1094, opened 2015-08-11
#1057 Use 5432 as the default port

derekjn:
@jberkus We're now using 5432 as the default port to avoid conflicts with existing libpq installations. I updated the Docker scripts to reflect this (we also removed the dependency on `libcurl`).


usmanm:
I'd say let's still keep the `\|`. A little perk of using `psql` that's installed with PipelineDB. We don't have to document it, but it's good while debugging.

derekjn:
Fair enough!

jberkus:
Merged,

In the process, I picked up a file I had omitted from my commit adding support for cstore in the docker package.

derekjn:
Thanks!

jberkus:
Feh, Docker build is blowing up now because the PipelineDB packages haven't been updated.

PR #1096, opened 2015-08-12
Major parser and analyzer overhaul

usmanm:


src/backend/pipeline/cont_analyze.c
```diff
@@ -278,136 +430,238 @@ find_clock_timestamp_expr(Node *node, ContAnalyzeContext *context)
...
+		node = lfirst(lc);
 
-	return NULL;
+		if (equal(node, expr))
+			continue;
+
+		if (contains_node(node, col))
+			ereport(ERROR,
+					(errcode(ERRCODE_SYNTAX_ERROR),
+					errmsg("the column being compared to clock_timestamp() cannot used in a GROUP BY expression "
```
derekjn: Rejecting this is for convenience in the analyzer, right? It seems like technically we could just remove this from the grouping, hoist it into the target list, and put it in the overlay's group clause.

I'm not saying we should do that, I'm just trying to understand why it's not allowed.

src/backend/pipeline/cont_analyze.c
```diff
@@ -966,72 +1289,1921 @@ transformContSelectStmt(ParseState *pstate, SelectStmt *select)
...
-		coldef->is_not_null = false;
-		coldef->raw_default = NULL;
-		coldef->cooked_default = NULL;
-		coldef->constraints = NIL;
-		coldef->typeName = typename;
+	arg = copyObject(expr);
 
-		stmt->base.tableElts = lappend(stmt->base.tableElts, coldef);
-	}
+	// FIXME(usmanm): Use replace_node.
```
derekjn: Is this relevant?

usmanm: Yes, I'll fix this.

src/backend/pipeline/cont_analyze.c
usmanm: So we allow simple grouping on the timestamp column, e.g. `SELECT arrival_timestamp FROM stream WHERE arrival_timestamp > clock_timestamp() - interval '5 second' GROUP BY arrival_timestamp`. This disallows something like grouping on `md5(arrival_timestamp)`.

derekjn: And that could obviously be supported pretty easily too, although it wouldn't be very sensible because it would basically require storing all rows and doing the grouping at read time. Got it!

derekjn:
I'm already pretty familiar with this branch but gave it a good once over. All of that red brings joy to my heart--this is beautiful, conscientious, top-notch work man.

Once tests are passing reliably, merge when ready.

PR #1100, opened 2015-08-13
Coarser grained default step size for sliding windows

usmanm:


src/backend/pipeline/cont_analyze.c
```diff
@@ -1886,20 +1890,58 @@ truncate_timestamp_field(Node *expr, ContAnalyzeContext *context)
...
+
+			expr = (Expr *) transformExpr(make_parsestate(NULL), ct_expr->rexpr, EXPR_KIND_WHERE);
+			Assert(IsA(expr, Const));
+			c = (Const *) expr;
+			Assert(c->consttype == INTERVALOID);
+
+			i = (Interval *) c->constvalue;
+
+			/* We make the step size one unit smaller than the granularity of the window size unit */
+			if (i->month >= 12)
```
derekjn: I think days should be the coarsest step size when automatically choosing. It might even make sense to make it hours. Only seeing updates once per month or day seems like the other extreme of the scale in terms of intuitive behavior that users expect.

derekjn:
Looks good, this should be a solid improvement for users. A couple of things:

* Can you add a test?
* Do you think we should log a `NOTICE` when we automatically select a step size? It could communicate to users how often to expect the data to update, and how to override this behavior if that's not what they want.

usmanm:
Tests automatically got updated with the notice--two birds with one stone.

derekjn:
Haha, nice!

Issue #1083, opened 2015-08-04
Display continuous view definition when describing a CV


derekjn:
Currently `\d+` on a CV will display the overlay view. `\d` should display the continuous view definition.

PR #1106, opened 2015-08-13
#1083 Display continuous view definition when describing a continuous…

derekjn:
`\d` on a continuous view will now show the CV's definition. Overlay views can be retrieved via:

    SELECT pipeline_get_overlay_viewdef('cv name')


usmanm:
Nice, looks good!

PR #1112, opened 2015-08-14
Fix user combines

usmanm:
For now it disallows combining expressions containing aggregates, but we can probably fix that later on if people ask for it.


derekjn:
Looks good man

PR #1114, opened 2015-08-15
Multiplicative back off for queries that throw errors 

usmanm:
Still trying to figure out how to best test this, ideas?


derekjn:
Don't we want this to also work if the process crashes even before entering the loop? If that happens the proc will enter a recovery loop that can potentially peg the CPU.

usmanm:
That control will probably have to go into the scheduler, because outside of this the process will crash with the postmaster will restart it.

derekjn:
We should just make the scheduler stop trying to restart a proc that crashes repeatedly. Is that outside the scope of this diff?

usmanm:
Yea, that'll require tweaking the background worker infrastructure, which I have been meaning to for a bit. Will create an issue for it.

derekjn:
Regarding testing, you could write a pytest that does something evil like deletes the matrel's files under the data directory which will surely cause the combiner to explode. Then you can probably look at the combiner stats and verify that the number of errors is reasonable (shouldn't be ridiculously high).

usmanm:
Going to create PLpgSQL aggregates that throw errors on some predefined inputs, and then supply those inputs and see how data is lost.

usmanm:
This is good to go now.

usmanm:
This is still unstable, so going to remove it from `0.8.0`.

usmanm:
Will re-open when ready.

Issue #1058, opened 2015-07-23
Single matrel for multiple sliding window queries


usmanm:
In cases where the only thing that is different is the window size

This ties into the general need to re-use existing matrels where possible.

derekjn:
I think it might make sense to just let users create views on top of SW CVs if they want to reuse matrels. There are legitimate reasons for wanting separate CVs (such as read latency). And as you mentioned @usmanm, this conflicts with #1120 which is probably more useful anyways.

PR #1140, opened 2015-09-05
#1058 Expose step column as arrival_timestamp in matrels to make it m…

derekjn:
Exposes the hidden step column as `arrival_timestamp`, which makes it more intuitive for users to create multiple windows over the same matrel using `combine`.


derekjn:
https://github.com/pipelinedb/pipelinedb/pull/1142

Issue #1120, opened 2015-08-17
Dynamic step sizes for sliding windows


usmanm:
Based on some configurable factor

usmanm:
This might conflict with #1058

derekjn:
It should be configurable at the CV level too. e.g.:

```
CREATE CONTINUOUS VIEW v WITH (sw_step_size = 1) AS -- seconds
    SELECT COUNT(*) FROM stream
WHERE (arrival_timestamp > clock_timestamp() - interval '1 hour');
```

or as fraction of the window size:

```
CREATE CONTINUOUS VIEW v WITH (sw_step_factor = 0.1) AS -- % of window size
    SELECT COUNT(*) FROM stream
WHERE (arrival_timestamp > clock_timestamp() - interval '1 hour');
```

usmanm:
Needs #1155

PR #1143, opened 2015-09-06
Sliding-window improvements

derekjn:
This actually addresses both #1120 and #1058.

Users can now create multiple sliding windows over the same CV by referencing `arrival_timestamp`, which gets pushed down to the matrel `SELECT` at rewrite time. e.g.:

```
CREATE CONTINUOUS VIEW sw0 AS SELECT COUNT(*) FROM stream
WHERE arrival_timestamp > clock_timestamp() - interval '1 hour';

CREATE VIEW sw1 AS SELECT combine(count) FROM sw0
WHERE arrival_timestamp > clock_timestamp() - interval '10 minutes';

CREATE VIEW sw1 AS SELECT * FROM sw0
WHERE arrival_timestamp > clock_timestamp() - interval '10 seconds';
```

I also ripped out the notice about window and step size that we were showing users upon sliding-window CV creation because it's sort of confusing with the step factor, and it made tests too finicky.


src/backend/pipeline/cont_analyze.c
```diff
@@ -484,28 +485,44 @@ find_clock_timestamp_expr(Node *node, ContAnalyzeContext *context)
...
+	min_step.time = 1000 * 1000;
 
 	Assert(IsA(expr, Const));
-	c = (Const *) expr;
-	Assert(c->consttype == INTERVALOID);
+	window = (Const *) expr;
+	Assert(window->consttype == INTERVALOID);
+
+	/* We make the step size some percentage of the total window width */
+	step = (Interval *) DirectFunctionCall2(interval_mul, window->constvalue, Float8GetDatum(factor));
```
usmanm: Not sure if this will always work since `date_floor` only works when the interval size spans a single unit. So if the step here was `INTERVAL '75 second'`, `date_floor` would fail.

src/backend/pipeline/cont_analyze.c
```diff
@@ -2183,11 +2183,14 @@ proj_and_group_for_windows(SelectStmt *proc, SelectStmt *view, ContAnalyzeContex
 	}
 
 	/* Create and truncate projection and grouping on temporal expression */
+	context->hoisted_name = ARRIVAL_TIMESTAMP;
```
usmanm: What if the query looks something like `SELECT x::timestamp AS arrival_timestamp FROM stream WHERE arrival_timestamp > clock_timestamp() - INTERVAL '5 second'`?

derekjn: For some reason I though we were rejecting `arrival_timestamp` as a column name or alias, which we probably should.

src/backend/pipeline/cont_analyze.c
```diff
@@ -584,23 +582,6 @@ validate_window_timestamp_expr(SelectStmt *stmt, Node *node, ContAnalyzeContext
...
-	{
-		A_Expr *ct_expr = (A_Expr *) context->expr;
-		TypeCast *c;
-		char *fname;
-
-		Assert(ct_expr && IsA(ct_expr, A_Expr));
-
-		c = (TypeCast *) ct_expr->rexpr;
-		Assert(IsA(c, TypeCast) && IsA(c->arg, A_Const));
-		fname = get_truncation_from_interval_expr((Node *) c);
```
usmanm: Are we not doing this at all now?

usmanm:
Looks good!

PR #1146, opened 2015-09-08
Kafka extension

derekjn:
Uses background workers to consume from a Kafka topic into a relation. Partitions are round robined across however many workers are started with `kafka_consume_begin`.


contrib/pipelinedb_kafka/README.md
```diff
@@ -0,0 +1,79 @@
...
+=# CREATE EXTENSION pipelinedb_kafka;
+CREATE EXTENSION
+```
+
+# Usage
+
+`pipelinedb_kafka` needs to know about any Kafka brokers it can connect to. Brokers can be added with `kafka_add_broker`:
+
+```
+=# SELECT kafka_add_broker('localhost:9092');
```
usmanm: Is there any `kafka_remove_broker` UDF as well?

contrib/pipelinedb_kafka/pipelinedb_kafka.c
```diff
@@ -0,0 +1,895 @@
...
+create_consumer(Relation consumers, text *relation, text *topic, text *format, text *delimiter, int batchsize)
+{
+	HeapTuple tup;
+	Datum values[CONSUMER_RELATION_NATTS];
+	bool nulls[CONSUMER_RELATION_NATTS];
+	Oid oid;
+
+	values[CONSUMER_ATTR_RELATION - 1] = PointerGetDatum(relation);
+	values[CONSUMER_ATTR_TOPIC - 1] = PointerGetDatum(topic);
+	values[CONSUMER_ATTR_PARTITION - 1] = Int32GetDatum(0);
```
usmanm: Should we expose allowing consumption from different partitions?

contrib/pipelinedb_kafka/pipelinedb_kafka.c
```diff
@@ -0,0 +1,895 @@
...
+	}
+
+	MemSet(nulls, false, sizeof(nulls));
+	MemSet(replace, false, sizeof(nulls));
+
+	replace[CONSUMER_ATTR_PARTITION - 1] = true;
+	replace[CONSUMER_ATTR_OFFSET - 1] = true;
+	replace[CONSUMER_ATTR_BATCH_SIZE - 1] = true;
+
+	values[CONSUMER_ATTR_PARTITION - 1] = Int32GetDatum(0);
```
usmanm: We probably only need to update the offset here?

contrib/pipelinedb_kafka/pipelinedb_kafka.c
```diff
@@ -0,0 +1,895 @@
...
+	desc = RelationGetDescr(rel);
+
+	for (i = 0; i < desc->natts; i++)
+	{
+		/*
+		 * Users can't supply values for arrival_timestamp, so make
+		 * sure we exclude it from the copy attr list
+		 */
+		char *name = NameStr(desc->attrs[i]->attname);
+		if (rel->rd_rel->relkind == RELKIND_STREAM && pg_strcasecmp(name, ARRIVAL_TIMESTAMP) == 0)
```
usmanm: So we allow writing to tables as well?

contrib/pipelinedb_kafka/README.md
derekjn: I'll add it.

contrib/pipelinedb_kafka/pipelinedb_kafka.c
derekjn: Yea that's the last thing I'm working on. The consumer proc will actually need to cycle through partitions after retrieving all of them from the Kafka cluster. I'll probably add some kind of `parallelism` arg to `kafka_consume_begin` so that each proc will read from `# partitions / # procs` partitions

contrib/pipelinedb_kafka/pipelinedb_kafka.c
derekjn: Yep, I don't see a reason to not allow it.

usmanm: Agreed.

usmanm:
Maybe we should keep this in a separate repo?

derekjn:
I don't think putting it into a separate repo would benefit users until we get the generic `COPY` hook into upstream so it can be used with PG. This extension is also going to be included with our binaries so that Kafka support works out of the box, so I think it makes sense to maintain consistency with how extensions are shipped with PG.

Anyways that was my original line of thinking. I'm happy to change things up if you think it would benefit users though.

usmanm:
No, that makes sense. It can't really work with vanilla Postgres for now so
not much value in keeping it as a separate repo for now.

On Saturday, September 12, 2015, Derek Nelson <notifications@github.com>
wrote:

> I don't think putting it into a separate repo would benefit users until we
> get the generic COPY hook into upstream so it can be used with PG. This
> extension is also going to be included with our binaries so that Kafka
> support works out of the box, so I think it makes sense to maintain
> consistency with how extensions are shipped with PG.
>
> Anyways that was my original line of thinking. I'm happy to change things
> up if you think it would benefit users though.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pipelinedb/pipelinedb/pull/1146#issuecomment-139807001>
> .
>


-- 
Usman

PR #1148, opened 2015-09-09
Fix group caching

usmanm:


src/backend/pipeline/cont_combiner.c
```diff
@@ -488,21 +494,26 @@ sync_combine(ContQueryCombinerState *state, Tuplestorestate *results, TupleHashT
...
+	foreach(lc, cache_tuples)
 	{
-		foreach_tuple(slot, results)
-		{
-			GroupCachePut(state->cache, slot);
-		}
+		HeapTuple tup = (HeapTuple) lfirst(lc);
+		ExecStoreTuple(tup, slot, InvalidBuffer, false);
+		GroupCachePut(state->cache, slot);
+		heap_freetuple(tup);
```
derekjn: This probably isn't necessary since the current `MemoryContext` is about to get deleted.

src/backend/pipeline/cont_combiner.c
```diff
@@ -488,21 +494,26 @@ sync_combine(ContQueryCombinerState *state, Tuplestorestate *results, TupleHashT
...
 	}
 
+	list_free(cache_tuples);
```
derekjn: Also not necessary.

derekjn:
Looks good. We should run some perf tests to see what this gets us...

src/backend/pipeline/cont_combiner.c
usmanm: I just try to free up any in-memory tuples eagerly since they can hog a lot of memory even if for a short duration.

derekjn: Cool, makes sense since `GroupCachePut` is copying everything.

PR #1157, opened 2015-09-11
Don't allow changing matrels directly

usmanm:


src/backend/utils/misc/pipelinedb.conf.sample
```diff
@@ -627,6 +627,16 @@
 # each database
 #continuous_query_num_workers = 2
 
+# allow direct changes to be made to materialization tables?
+#continuous_query_materialization_table_updatable = off
```
derekjn: Let's leave it undocumented for now.

src/backend/executor/execMain.c
```diff
@@ -968,6 +970,20 @@ CheckValidResultRel(Relation resultRel, CmdType operation)
 	switch (resultRel->rd_rel->relkind)
 	{
 		case RELKIND_RELATION:
+			if (!MatRelUpdatesEnabled())
```
derekjn: This is pedantic, but we might as well make this a macro.

derekjn:
Looks good

src/backend/utils/misc/pipelinedb.conf.sample
usmanm: Any reason to do so? Only power users ever really look at configuration options and it's prolly ok to expose this to them?

derekjn: Nope, feel free to leave it documented if you'd prefer.

PR #1159, opened 2015-09-12
Test for SW vacuum

usmanm:


derekjn:
Nice man, looks good!

Issue #1144, opened 2015-09-07
Add SW convenience function


derekjn:
e.g. instead of:

```
CREATE CONTINUOUS VIEW v AS SELECT COUNT(*)
FROM stream WHERE arrival_timestamp > clock_timestamp() - interval '1 minute';
```
something like:
```
CREATE CONTINUOUS VIEW v AS SELECT COUNT(*)
FROM stream WHERE arrived_within(interval '1 minute');
```
or even:
```
CREATE CONTINUOUS VIEW v AS SELECT COUNT(*)
FROM stream WHERE arrived_within('1 minute');
```
or both...or even a way to conceptualize a window as a TTL:

```
CREATE CONTINUOUS VIEW v WITH (ttl=60) AS SELECT COUNT(*) FROM stream;
CREATE CONTINUOUS VIEW v WITH (max_age=60) AS SELECT COUNT(*) FROM stream;
```

PR #1161, opened 2015-09-13
#1144 Add SW convenience parameter (max_age)

derekjn:
SWs can now be created using a `WITH` param:

```
CREATE CONTINUOUS VIEW cv WITH (max_age = '10 days') AS SELECT count(*) FROM stream
CREATE VIEW v WITH (max_age = '1 day') AS SELECT * from cv;
```


usmanm:
Looks good! This is probably a more intuitive way of thinking about sliding windows.

PR #1171, opened 2015-09-15
Add date_round_down

usmanm:


src/backend/utils/adt/timestamp.c
```diff
@@ -4170,12 +4170,11 @@ Datum
...
 	TimestampTz timestamp = PG_GETARG_TIMESTAMPTZ(0);
-	Interval *span = PG_GETARG_INTERVAL_P(1);
-	fsec_t fsec;
-	struct pg_tm i_tm;
-	struct pg_tm ts_tm;
-	int tz;
-	int nset;
+	TimestampTz interval_ts;
+
+#ifndef HAVE_INT64_TIMESTAMP
```
usmanm: Figured this is kosher to do since all modern machines support 64-bit integers. If someone doesn't have them, then they're in bad luck.

src/test/regress/expected/date_floor.out
```diff
@@ -33,35 +33,55 @@ SELECT date_floor(TIMESTAMP '2001-02-16 20:38:40', INTERVAL '10 minute');
 SELECT date_floor(TIMESTAMP '2001-02-16 20:38:40', INTERVAL '3 hour');
           date_floor          
 ------------------------------
- Fri Feb 16 18:00:00 2001 PST
+ Fri Feb 16 19:00:00 2001 PST
```
derekjn: I'm a bit confused by this. Intuitively it seems that the floor should be a multiple of the interval size, right?

loadzero: Is this a DST thing?

usmanm: We should treat this function as an internal function. Maybe I should rename it. It's only purpose is to create fixed size time buckets of duration equal to the interval specified.

derekjn: Ok, that makes sense.

PR #1173, opened 2015-09-15
SW views can be created even if time column alias isn't arrival_timestamp

usmanm:
Also, since we allow SWs to be defined on columns other than `arrival_timestamp`, maybe we should use an alias like `sw_column` to refer to the time column in the SW clause rather than `arrival_timestamp` when referring to it in `CREATE VIEW`.

### DIFF
--- a/src/backend/catalog/dependency.c
+++ b/src/backend/catalog/dependency.c
@@ -1608,7 +1608,7 @@ find_expr_references_walker(Node *node,
 		List	   *rtable;
 		RangeTblEntry *rte;
 
-		if (IS_ARRIVAL_TIMESTAMP_REF(var))
+		if (IS_SW_TIMESTAMP_REF(var))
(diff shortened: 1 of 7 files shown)


derekjn:
Looks good!

usmanm:
Should we use the alias sw_timestamp instead of arrival_timestamp to refer to the column?

derekjn:
I don't know of any of our users who are using sliding windows without `arrival_timestamp`, so I think that changing the alias `sw_timestamp` may be confusing to users. The correct way to handle this is probably to actually detect the name of the underlying sliding window column and allow that to be pushed down. That may require a bit of work and probably isn't worth spending time on just yet.

usmanm:
I think it shouldn't be too tricky to go. Lemme see if I can crank it out in an hour.

PR #1184, opened 2015-09-24
Shard on group

derekjn:
Grouped queries are now sharded across combiners by the hash of their group. Some notes on how it works:

* Workers now attach a hash value to each `TupleBufferSlot` that is written to the `CombinerTupleBuffer`. This is the same hash that the matrel index computes, so some hashing overhead is now offloaded to worker procs.
* For non-grouped queries, the continuous query id is simply used as the hash value, so these types of queries are effectively sharded by id like we've been doing.


src/backend/pipeline/cont_combiner.c
```diff
@@ -81,8 +76,18 @@ typedef struct {
 	Relation matrel;
 	ResultRelInfo *ri;
 	CQStatEntry stats;
+	TupleHashTable existing;
 } ContQueryCombinerState;
 
+/* current query state being used in ContinuousQueryCombinerMain */
+static ContQueryCombinerState *current_state;
```
usmanm: Don't think this is needed anymore?

src/backend/pipeline/cont_combiner.c
```diff
@@ -812,9 +786,18 @@ read_batch(ContQueryCombinerState *state, TupleBufferBatchReader *reader)
...
 
 		IncrementCQRead(1, tbs->size);
 		count++;
+
+		/*
+		 * This is fairly unlikely, but if it does happen we should
+		 * finish processing this batch immediately because we're storing 8-byte hashes
+		 * in memory alongside each of these tuples.
+		 */
+		if (!tuplestore_in_memory(state->batch))
```
usmanm: Not sure if I follow what's going on here? We're already allocating `batch_size * 8` bytes for each state. Whether the batch of tuples fits in memory or not shouldn't make a difference?

src/backend/pipeline/cont_combiner.c
```diff
@@ -938,12 +917,22 @@ ContinuousQueryCombinerMain(void)
...
+				 */
+				if (state->ngroupatts == 0 && id % continuous_query_num_combiners != MyContQueryProc->group_id)
+					continue;
+
 				debug_query_string = NameStr(state->view->name);
 				MemoryContextSwitchTo(state->tmp_cxt);
 
 				TupleBufferBatchReaderSetCQId(reader, id);
 
+				/* give should_read_fn access to the current state */
```
usmanm: This isn't needed either.

src/backend/pipeline/cont_combiner.c
```diff
@@ -812,9 +786,18 @@ read_batch(ContQueryCombinerState *state, TupleBufferBatchReader *reader)
 
 		ExecStoreTuple(heap_copytuple(tbs->tuple->heaptup), state->slot, InvalidBuffer, false);
 		tuplestore_puttupleslot(state->batch, state->slot);
+		state->batch_hashes[count] = tbs->tuple->group_hash;
```
usmanm: We can probably avoid allocating this large array for each state by keeping a single global one and re-using it?

src/backend/pipeline/cont_combiner.c
derekjn: Whoops!

src/backend/pipeline/cont_combiner.c
derekjn: This is actually not even really related to this PR, but more of a general corner case. The comment doesn't even need to mention the hashes.

Once the initial memory allocated for a tuplestore has been exceeded, new tuples will start spilling to disk. It's better just to finish the batch immediately to avoid that. With the current settings, it would only take 10k ~3KB tuples to cause this to happen, which isn't likely but certainly possible.

I can put this change in a separate diff, like I said I just happened to notice it as I was reading the code.

usmanm: Isn't the combiner working memory = 256mb? So it should take way more than that tuples for this tuple store to fill up?

The comment sounds misleading though, perhaps we should just say that never spill batches to disk.

src/backend/pipeline/cont_combiner.c
derekjn: Good idea

src/backend/pipeline/cont_combiner.c
derekjn: Ah you're right, I thought we were only using 32mb for the batch tuplestore. Even then it's within the realm of possible (10's of thousands of tuples x 10's of KB each), but at that point perf characteristics would be completely unpredictable anyways.

It still seems like an easy thing to avoid though, but I can revert this part if you'd like. I also updated the comment.

usmanm: No, keep it. I'm now wondering if we should stop when memory is <= half of 256mb (or whatever the combiner working memory is), because in worst case we could look up exactly the number of tuples from the matrel as there exist in the batch?

Moreover, the size of the tuple buffer is only 64mb which seems too low compared to the 256mb working memory for the batch being read from it?

PR #1193, opened 2015-09-29
Fixes #1167: catch type resolution issue at CV create time

### LINKED ISSUE #1167: array_agg error
`CREATE CONTINUOUS VIEW v AS SELECT array_agg(ARRAY[x::int, y::int]) FROM stream;` doesn't give an error at creation time but when trying to insert into the stream, we see an error:

```
WARNING:  snapshot 0x2d2ed30 still active
ERROR:  could not find array type for data type integer
```

derekjn:


usmanm:
Looks good!

PR #1197, opened 2015-09-29
Dynamic step size for sliding windows

usmanm:


src/backend/catalog/pipeline_query.c
```diff
@@ -342,33 +342,35 @@ GetCVNameFromMatRelName(RangeVar *name)
...
+SelectStmt *
+GetContSelectStmt(RangeVar *rv)
 {
 	HeapTuple tuple;
 	bool isnull;
 	Datum tmp;
-	char *result = NULL;
+	char *sql;
+	Query *query;
+	SelectStmt *select;
```
derekjn: We use the same variable name in `cont_analyze.c` but we should probably stop using it since it's also a syscall...

src/backend/utils/misc/guc.c
```diff
@@ -2668,6 +2669,16 @@ static struct config_int ConfigureNamesInt[] =
 		NULL, NULL, NULL
 	},
 
+	{
+		{"sliding_window_step_factor", PGC_USERSET, QUERY_TUNING_OTHER,
+		 gettext_noop("Steps the default step size for a sliding window query as a factor of the window size."),
```
derekjn: ~~Steps t~~The default step size for a sliding window query as a factor of the window size.

derekjn:
Looks good! As an aside, we should start thinking about more flexible/less intrusive ways to attach state to queries without having to add fields to primitive nodes.

usmanm:
Maybe I should add a JSON field to `pipeline_query`?

derekjn:
Maybe we could go even more generic with a `pipeline_config` table with a schema like: (namespace oid, key text, value bytea)?

usmanm:
We have `pipeline_tstate`, maybe we should utilize that?

derekjn:
That would actually work because `distinct` is a `varlena` but that seems like a pretty awful hack. Definitely go ahead and merge this and lets talk about the generic state storage at some point.

usmanm:
We can probably create a new `jsonb` field in `pipeline_tstate` and rename the catalog to something a little more sane.

PR #1199, opened 2015-09-30
1078 adhoc backend

loadzero:
Adhoc query support (sliding windows not implemented yet)

To try it out, run 
```
padhoc 'select key::text, count(*) from stream group by key'
```

and then insert some rows

```
bin/generate-inserts --stream stream  --key=str --value=int --batchsize=100 --n=1000  | psql
```

You should see the query result live updating in the padhoc client.

If you execute from inside psql, you can see the adhoc messages streaming in copy format

```
psql -c 'select key::text, count(*) from stream group by key'
```

Issues.

Sliding windows not supported.

IPC for adhoc queries needs looking at, I have currently piggy backed on the mechanism for bg workers, but it is a bit of a hack. see cont_scheduler.c:start_group and 
cont_adhoc.c:get_cont_query_proc

Perhaps adhoc queries should get their own tuple buffer?


derekjn:
We need to add the ncurses development package as a dependency. I didn't have ncurses.h on my system and couldn't compile initially.

usmanm:
I get the error:
```
usmanm=# INSERT INTO stream (key) VALUES ('usmanm');
ERROR:  relation "stream" does not exist
LINE 1: INSERT INTO stream (key) VALUES ('usmanm');
```

You probably have to commit the transaction that creates the adhoc continuous query so it's visible to other backends.

derekjn:
I tried to run `padhoc 'select key::text, count(*) from stream group by x'` and got this:

    could not execute query select key::text, count(*) from stream group by x

And on the server logs:

```
ERROR:  column reference "x" has an ambiguous type at character 49
HINT:  Explicitly cast to the desired type, e.g. x::integer, or create the stream "stream" with CREATE STREAM.
```

loadzero:
Damn, that was a typo, should be

```
./src/bin/psql/padhoc 'select key::text,count(*) from stream group by key'
```

derekjn:
Currently `padhoc` hides error messages with `Currently they appear to overridden with...`. It should definitely propagate error messages to the client the same way that `psql` does.

src/bin/psql/adhoc.c
```diff
@@ -51,7 +50,13 @@ main(int argc, char *argv[])
 	Model *model = ModelInit();
 
 	Screen *screen = 0;
-	RowStream *stream = 0;
+	PGRowStream *stream = 0;
+
+	if (argc < 2)
+	{
+		fprintf(stderr, "usage: padhoc sql\n");
```
derekjn: `padhoc` should mimic the connection options of `psql`:

```
Connection options:
  -h, --host=HOSTNAME      database server host or socket directory (default: "local socket")
  -p, --port=PORT          database server port (default: "5432")
  -U, --username=USERNAME  database user name (default: "derek")
  -w, --no-password        never prompt for password
  -W, --password           force password prompt (should happen automatically)
```

src/backend/pipeline/cont_scheduler.c
```diff
@@ -46,7 +46,9 @@
 
 #define MAX_PROC_TABLE_SZ 16 /* an entry exists per database */
 #define INIT_PROC_TABLE_SZ 4
-#define TOTAL_SLOTS (continuous_query_num_workers + continuous_query_num_combiners)
+
+/* extra +1 slot is for the adhoc query */
+#define TOTAL_SLOTS (continuous_query_num_workers + continuous_query_num_combiners + 1)
```
derekjn: We'll want an adhoc backend process for each individual adhoc query.

derekjn: It seems that the only use an adhoc proc has for the existing process metadata infrastructure is to read tuples the same way that continuous views currently do. We'll probably need to rework that slightly so that adhocs procs can't interfere with regular combiner/worker procs, so we should be able to just do away with any changes to this file.

src/backend/tcop/postgres.c
```diff
@@ -988,6 +990,12 @@ exec_simple_query(const char *query_string)
 			break;
 		}
 
+		if (IsA(parsetree, SelectStmt) && IsAdhocQuery(parsetree))
```
derekjn: Obviously this can interfere quite a bit with the standard execution path. Perhaps the best way to initialize an adhoc CQ would be with a function call:

    SELECT exec_adhoc_continuous_query('SELECT ... FROM ...');

This would avoid having to change any syntax and would make adhoc detection reliable, as everything would be isolated within the body of a single function.

src/backend/pipeline/cont_plan.c
```diff
@@ -119,6 +119,45 @@ get_worker_plan(ContinuousView *view)
 	return get_plan_from_stmt(view->id, (Node *) selectstmt, view->query, false);
 }
 
+PlannedStmt*
+GetContinuousViewPlan(ContinuousView *view)
```
derekjn: We call what's being retrieved here the "overlay" view, so this should probably be named something like `GetContinuousViewOverlayPlan`.

Additionally, this is basically doing the same work that retrieving the combiner plan requires, which isn't necessary for getting the overlay plan. It's basically identical to getting the worker plan, but with passing a pointer to `TransformSelectStmtForContProcess`:

```
    TransformSelectStmtForContProcess(view->matrel, selectstmt, &overlay_stmt, Worker);
    return get_plan_from_stmt(view->id, (Node *) overlay_stmt, view->query, false);
```

src/backend/pipeline/cont_adhoc_sender.c
```diff
@@ -0,0 +1,373 @@
...
+	int numCols;
+};
+
+/* send the buffered message in copy format, and flush.
+ * caller is responsible for clearing the buffer */
+
+static void
+adhoc_dest_send(struct AdhocDestReceiver *adhoc_dest)
+{
+	pq_putmessage('d', adhoc_dest->msg_buf->data, adhoc_dest->msg_buf->len);
```
derekjn: After reading through the logical decoding code, your decision to use a simple protocol wrapped in `COPY` makes a lot of sense. Logical decoding essentially provides a callback interface to do something with an `XLog` record when one is created. There's no standard textual representation as far as I can tell.

Actual continuous view updates will already work with logical decoding since they're WAL-logged, but it's worth thinking about how we can hook adhoc CQs in to the logical decoding infrastructure. It's not that important for the time being though.

derekjn:
We should also set the priority of adhoc procs to something relatively low. Since they're fully in memory, they can easily eat up a lot of CPU.

derekjn:
To summarize the outstanding issues so we can just cross them off here:

* Enter adhoc loop via a function call
* TupleBuffer reading rework
* `padhoc` error propagation
* `padhoc` connection options
* `GetContinuousViewOverlayPlan`
* `nice`

src/backend/pipeline/cont_analyze.c
```diff
@@ -1560,7 +1560,7 @@ transformContSelectTargetList(ParseState *pstate, List *tlist)
 				 * CV's aggregate types will not have been changed. This can cause a
 				 * signature resolution error when resolving the transout function.
 				 */
-				if (IsContQueryProcess())
+				if (IsContQueryProcess() || IsContQueryAdhocProcess())
```
usmanm: Maybe we should just change the `IsContQueryProcess()` macro to return true for ad-hoc procs as well?

src/backend/pipeline/cont_plan.c
usmanm: You can probably just re-use `GetContPlan(view, Combiner)`.

src/backend/tcop/postgres.c
usmanm: +1

PR #1214, opened 2015-10-14
1211 topk byref

derekjn:


src/backend/pipeline/fss.c
```diff
@@ -55,19 +45,33 @@ FSSCreateWithMAndH(uint64_t k, TypeCacheEntry *typ, uint64_t m, uint64_t h)
 
 	if (!typ->typbyval)
 	{
+		int i;
+
+		/* For byref types, we use the value as an index into the varlena array */
+		for (i = 0; i < m; i++)
+			fss->monitored_elements[i].value = (Datum ) i;
```
usmanm: Instead of this, we should store their Murmur hash and use that as a faster equality check.

src/backend/pipeline/fss.c
```diff
@@ -87,12 +107,63 @@ FSSDestroy(FSS *fss)
...
+ */
+static Datum
+get_monitored_value(FSS *fss, MonitoredElement *element, bool *isnull)
+{
+	Datum result = element->value;
+	int index;
+
+	*isnull = false;
+
+	if (fss->typ.typbyval)
```
usmanm: What about null check for by val types? Seems like we'll always return `isnull = false`.

usmanm:
Looks good!

PR #1218, opened 2015-10-15
1078 adhoc backend

loadzero:
This is PR #1199 with a bunch of issues fixed:

* TupleBuffer reading rework
* Enter adhoc loop via a function call
* padhoc error propagation
* padhoc connection options
* GetContinuousViewOverlayPlan
* niceness
* Style-guide conformity

The bulk of the work has gone into setting up a new tuple buffer for 
adhoc queries (AdhocTupleBuffer), and dealing with synchronization issues.

Particular things to pay attention to:

AdhocMgrPeriodicCleanup and its caller (cont_scheduler.c)
AdhocMgrDeleteAdhocs and its caller (cont_scheduler.c)

streamReceiver.c and stream.c have been modified to handle copying tuples to
the adhoc tuple buffer.

tuplebuf.c has been modified in a couple of places for synchronization reasons.


usmanm:
Awesome, will review today!

usmanm:
@loadzero: Can we add a test where multiple continuous views and adhoc queries are reading from the same stream and see that everything works as expected?

derekjn:
@loadzero Can we move the entire `padhoc` build into `pipeline/bin/`? I don't see a reason to leave it in with `psql`.

loadzero:
@usmanm yeah, that kind of test makes sense - will add.
@derekjn I agree with moving it out, it does share some code with psql, but it should be possible.

derekjn:
Another test we should have is to create an adhoc query identical to a CV's query, write a bunch of random data to the stream they both read from, and then verify that their results are identical. As long as floating-point data isn't used this should be reliable.

derekjn:
`expect` needs to be added as a build dependency.

usmanm:
IT'S GREEN!

src/backend/catalog/pipeline_query.c
```diff
@@ -550,6 +551,12 @@ GetAllContinuousViewIds(void)
 	{
 		Form_pipeline_query row = (Form_pipeline_query) GETSTRUCT(tup);
 		Oid id = row->id;
+
+		if (row->adhoc)
+		{
```
usmanm: No need for braces.

src/backend/catalog/pipeline_stream.c
```diff
@@ -492,6 +493,39 @@ GetAllStreamReaders(Oid relid)
...
+	HeapTuple tup;
+	Bitmapset *result = NULL;
+
+	while ((tup = heap_getnext(scan_desc, ForwardScanDirection)) != NULL)
+	{
+		Form_pipeline_query row = (Form_pipeline_query) GETSTRUCT(tup);
+		Oid id = row->id;
+
+		if (!row->adhoc)
+		{
```
usmanm: Same, no braces if there's just one statement in the block.

src/backend/executor/nodeAgg.c
```diff
@@ -1814,7 +1814,7 @@ ExecInitAgg(Agg *node, EState *estate, int eflags)
 		peraggstate->transfn_oid = transfn_oid = aggform->aggtransfn;
 		peraggstate->finalfn_oid = finalfn_oid = aggform->aggfinalfn;
 
-		peraggstate->finalize = !(AGGKIND_IS_COMBINE(aggref->aggkind) || IsContQueryProcess());
+		peraggstate->finalize = !(AGGKIND_IS_COMBINE(aggref->aggkind) || IsContQueryProcess() || IsContQueryAdhocProcess());
```
usmanm: Should we maybe change the `IsContQueryProcess()` macro to return true for ad-hoc processes as well?

loadzero: Done. I had to change the assert inside make_finalize_for_viewdef though

```
@@ -2896,7 +2896,7 @@ make_finalize_for_viewdef(ParseState *pstate, RangeVar *cvrv, Var *var, Node *ar
        Oid finaltype = InvalidOid;
        Query *q = GetContWorkerQuery(cvrv);
 
-       Assert(!IsContQueryProcess());
+       Assert(IsContQueryAdhocProcess() || !IsContQueryProcess());
```

src/backend/catalog/pipeline_query.c
loadzero: fixed

src/backend/catalog/pipeline_stream.c
loadzero: fixed

loadzero:
@derekjn padhoc moived from src/bin/psql -> src/bin/pipeline as of e5b32fa
@usmanm multi cq / adhoc test harness added to test_adhoc.py as of 62bc13

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+	ContQueryBatchContext = AllocSetContextCreate(CurrentMemoryContext,
+			"ContQueryBatchContext",
+			ALLOCSET_DEFAULT_MINSIZE,
+			ALLOCSET_DEFAULT_INITSIZE,
+			ALLOCSET_DEFAULT_MAXSIZE);
+}
+
+/* util func to set reader on stream scan nodes */ 
+static void
+set_reader(PlanState *planstate, TupleBufferBatchReader *reader)
```
usmanm: This is identical to `set_reader` in `cont_worker.c`? Maybe we should move it to `cont_plan.c` and make it a  public function?

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+acquire_cont_query_proc()
+{
+	MyContQueryProc = AdhocMgrGetProc();
+}
+
+static void
+release_cont_query_proc()
+{
+	AdhocMgrReleaseProc(MyContQueryProc);
+	MyContQueryProc = 0;
```
usmanm: `MyContQueryProc = NULL`

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+		set_reader(((SubqueryScanState *) planstate)->subplan, reader);
+		return;
+	}
+
+	set_reader(planstate->lefttree, reader);
+	set_reader(planstate->righttree, reader);
+}
+
+static EState *
+create_estate(QueryDesc *query_desc)
```
usmanm: Same for these functions, not sure what the best place to put them would be though.

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+	if (!has_tup)
+	{
+		TupleBufferBatchReaderRewind(state->reader);
+		return false;
+	}
+
+	state->query_desc->estate = create_estate(state->query_desc);
+	estate = state->query_desc->estate;
+
+	TupleBufferBatchReaderSetCQId(state->reader, state->view_id);
```
usmanm: This only needs to be called once.

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+ */
+static bool
+exec_adhoc_worker(AdhocWorkerState * state)
+{
+	struct Plan *plan = 0;
+	bool has_tup = false;
+	EState *estate = 0;
+	ResourceOwner owner = CurrentResourceOwner;
+
+	TupleBufferBatchReaderTrySleepTimeout(state->reader, state->last_processed, 1000);
```
usmanm: Let's turn 1000 into a macro.

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+	TupleBufferBatchReaderSetCQId(state->reader, state->view_id);
+
+	set_snapshot(estate, owner);
+
+	plan = state->query_desc->plannedstmt->planTree;
+
+	state->query_desc->planstate = 
+		ExecInitNode(plan, state->query_desc->estate, 0);
+
+	set_reader(state->query_desc->planstate, state->reader);
```
usmanm: This only needs to be set once.

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+
+	ExecutePlan(estate, state->query_desc->planstate, 
+				state->query_desc->operation,
+				true, 0, 0, ForwardScanDirection, state->dest);
+
+	ExecEndNode(state->query_desc->planstate);
+	state->query_desc->planstate = NULL;
+	state->num_processed = estate->es_processed + estate->es_filtered;
```
usmanm: Don't think we need a field in the state struct. We can prolly just return `estate->es_processed + estate->es_filtered != 0` at the end.

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+		{
+			HeapTupleEntry entry = (HeapTupleEntry) 
+				LookupTupleHashEntry(state->existing, state->slot, NULL);
+
+			if (entry)
+			{
+				/* we are using 'result' as scratch space here. */
+				tuplestore_puttuple(state->result, entry->tuple);
+			}
+			else
```
usmanm: Remove empty else block.

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+			{
+			}
+		}
+
+		foreach_tuple(state->slot, state->result)
+		{
```
usmanm: No need for braces.

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+			old_cxt = MemoryContextSwitchTo(state->existing->tablecxt);
+			entry->tuple = ExecCopySlotTuple(state->slot);
+			MemoryContextSwitchTo(old_cxt);
+		}
+	}
+}
+
+/* get the view plan and set up the view state for execution */
+static AdhocViewState *
+init_adhoc_view(ContinuousViewData data,
```
usmanm: Should prolly pass a pointer for `ContinuousViewData`?

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+/* Execute a drop view statement to cleanup the adhoc view */
+void AdhocMgrCleanupContinuousView(ContinuousView *view)
+{
+	DestReceiver *receiver = 0;
+	DropStmt *stmt = makeNode(DropStmt);
+	List *querytree_list;
+	Node *plan;
+	Portal portal;
+
+	stmt->objects = list_make1(list_make1(makeString(view->name.data)));
```
usmanm: Use `NameStr(view->name)` instead of `(view->name.data)`.

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+	view_state->slot = combiner_state->slot;
+
+	CommitTransactionCommand();
+
+	cleanup_data.cont_view_data = &view_data;
+	cleanup_data.worker_state = worker_state;
+
+	PG_ENSURE_ERROR_CLEANUP(cleanup, PointerGetDatum(&cleanup_data));
+	{
+		while (true)
```
usmanm: Before getting into this loop we should set `XactReadOnly = true` since we don't perform any writes, but reset it to true afterwards so we can do the clean up.

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+	target = (ResTarget *) (first_node);
+	func = (FuncCall *) (target->val);
+	sql_arg = linitial(func->args);
+
+	return sql_arg->val.val.str;
+}
+
+/* set a lower process priority for adhoc backends */
+static void
+set_nice_priority()
```
usmanm: Let's also make this into a public function and share with `cont_scheduler.c`.

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+{
+	int default_priority = getpriority(PRIO_PROCESS, MyProcPid);
+	int priority = Max(default_priority, 20 - ceil(continuous_query_proc_priority * (20 - default_priority)));
+
+	priority = nice(priority);
+}
+
+static void
+set_default_priority()
```
usmanm: Store the default priority as a static variable before calling `set_nice_property`? Don't think it's too important to reset it, since this is probably called when we're exiting from the client?

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+			MemoryContextResetAndDeleteChildren(adhoc_cxt);
+			PG_RE_THROW();
+		}
+		PG_END_TRY();
+	}
+}
+
+/* returns true iff the function name matches the argument */
+static bool
+check_adhoc_func(FuncCall *func, const char *name)
```
usmanm: You don't need to pass `name` as an arg. Just create a macro for `"pipeline_exec_adhoc_query"`. Oh, and `NameStr` for everything where you're doing `NameData.data`.

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+
+	if (!IsA(sql_arg, A_Const))
+		return false;
+
+	return true;
+}
+
+/* returns true iff the query is an adhoc select */
+bool
+IsAdhocQuery(Node *node)
```
usmanm: Why do we need this? We should just register a function in `pg_proc.h` for `pipeline_exec_adhoc_query`. That way we don't need to add any checks to the query execution path and PG will simply drop is in that function's handler whenever `SELECT pipeline_exec_adhoc_query(...);` is called.

src/backend/pipeline/cont_adhoc.c
```diff
@@ -0,0 +1,1000 @@
...
+	return true;
+}
+
+/* 
+ * Placeholder function to prevent people from defining this function.
+ * This function is meant to be intercepted, not executed, hence the elog
+ * error if it does get executed.
+ */
+Datum
+pipeline_exec_adhoc_query(PG_FUNCTION_ARGS)
```
usmanm: Hm, why does this need to be intercepted and not executed?

src/backend/pipeline/cont_adhoc_mgr.c
```diff
@@ -0,0 +1,205 @@
...
+
+	if (grp)
+	{
+		MemSet(grp, 0, sizeof(ContQueryProcGroup));
+		grp->db_oid = MyDatabaseId;
+		grp->active = true;
+
+		proc = grp->procs;
+
+		proc->type = Scheduler;
```
usmanm: Not sure what's going on here. Why is this proc marked as a `Scheduler` proc? We should probably add a new enum for `Adhoc`?

src/backend/pipeline/cont_adhoc_sender.c
```diff
@@ -0,0 +1,373 @@
...
+		bool		isvarlena;
+
+		getTypeOutputInfo(attr[attnum - 1]->atttypid,
+					  &out_func_oid,
+					  &isvarlena);
+		fmgr_info(out_func_oid, &adhoc_dest->out_funcs[attnum - 1]);
+	}
+
+	if (!is_agg)
+	{
```
usmanm: No braces.

src/backend/pipeline/cont_plan.c
```diff
@@ -103,8 +103,8 @@ get_plan_from_stmt(Oid id, Node *node, const char *sql, bool is_combine)
 	return plan;
 }
 
-static PlannedStmt*
-get_worker_plan(ContinuousView *view)
+static SelectStmt* get_worker_select_stmts(ContinuousView* view,
```
usmanm: `get_worker_select_stmt`

src/backend/pipeline/cont_plan.c
```diff
@@ -114,9 +114,51 @@ get_worker_plan(ContinuousView *view)
...
 	selectstmt->swStepFactor = view->sw_step_factor;
-	selectstmt = TransformSelectStmtForContProcess(view->matrel, selectstmt, NULL, Worker);
+	selectstmt = TransformSelectStmtForContProcess(view->matrel, selectstmt, viewptr, Worker);
+
+	return selectstmt;
+}
 
-	return get_plan_from_stmt(view->id, (Node *) selectstmt, view->query, false);
+static PlannedStmt*
+get_worker_plan(ContinuousView *view, SelectStmt** viewptr)
```
usmanm: Don't think we need the `viewptr` argument here, since it's not used anywhere.

src/backend/pipeline/cont_scheduler.c
```diff
@@ -409,6 +423,10 @@ cq_bgproc_main(Datum arg)
 		break;
 	case Worker:
 		am_cont_worker = true;
+
+		/* XXX - put this somewhere nicer */
+		AdhocMgrDeleteAdhocs();
```
usmanm: Why is it here? Can't we put this in the start up of `cont_scheduler`? Or does this require a connection to a database?

src/backend/pipeline/cont_scheduler.c
```diff
@@ -703,12 +720,18 @@ ContQuerySchedulerMain(int argc, char *argv[])
...
 		 * signal handlers will wake us by calling SetLatch).
 		 */
-		rc = WaitLatch(&MyProc->procLatch, WL_LATCH_SET | WL_POSTMASTER_DEATH, 0);
+		rc = WaitLatch(&MyProc->procLatch, WL_LATCH_SET | WL_POSTMASTER_DEATH | WL_TIMEOUT, 1000);
 
 		ResetLatch(&MyProc->procLatch);
 
 		DisableCatchupInterrupt();
 
+		if (rc & WL_TIMEOUT)
```
usmanm: This place is probably fine. What happens if this clean up doesn't happen? Other than redundant entries in `pipeline_query`?

src/backend/pipeline/stream.c
```diff
@@ -248,6 +249,92 @@ DropPreparedStreamInsert(const char *name)
...
+	{
+		AdhocQuery *query = &adhoc_data->queries[i];
+
+		StreamTuple *tuple = 
+			MakeStreamTuple(tup, desc, 1, &query->ack);
+		Bitmapset *single = bms_make_singleton(query->cq_id);
+
+		tuple->group_hash = query->cq_id;
+
+		if (TupleBufferInsert(AdhocTupleBuffer, tuple, single))
```
usmanm: So there's one `AdhocTupleBuffer` for all ad-hoc processes?

src/backend/pipeline/stream.c
```diff
@@ -257,20 +344,34 @@ InsertIntoStreamPrepared(PreparedStreamInsertStmt *pstmt)
...
+
+	int num_worker = bms_num_members(targets);
+
 	InsertBatchAck acks[1];
 	InsertBatch *batch = NULL;
 	int num_batches = 0;
 	Size size = 0;
-	Bitmapset *targets = get_stream_readers(pstmt->relid);
+
+	init_adhoc_data(&adhoc_data, adhoc_targets);
```
usmanm: Should we only do all of this if `adhoc_targets` is not empty?

src/backend/pipeline/stream.c
```diff
@@ -315,7 +434,17 @@ InsertIntoStreamPrepared(PreparedStreamInsertStmt *pstmt)
 	stream_stat_report(pstmt->relid, count, 1, size);
 
 	if (synchronous_stream_insert)
```
usmanm: Can we try to unify this waiting code into some refactored functions? I feel like this whole sync stream stuff is getting a little ugly with this change. Lemme think about how we can clean it up.

src/bin/pipeline/rowmap.c
```diff
@@ -192,17 +186,42 @@ row_cmp_row_key(const void *p1, const void *p2)
...
-static void
-row_map_sort(RowMap *m)
+static int
+compare_key(void *ctx, const void *a, const void *b)
+{
+	const TreeNode *na = (const TreeNode *)(a);
+	const Key *nb = (const Key *)(b);
+
+	if (nb->is_key)
+	{
```
usmanm: No braces.

usmanm:
Looks good man! The only thing which looks a bit weird is the `InsertBatch` stuff. Lemme think of someway where it feels less of a sledge-hammer way to do it. I think it's fine to make that change in a separate diff, and merge this without the change for it. It's already too big of a change and slapping on more changes is prolly not the best idea. The branch looks like it's stable and is passing tests, which is awesome! One thing which might be of use is adding a config param to enable/disable ad-hoc queries. We can keep that as off by default till we're sure of the stability of this patch. 

Oh, and have you done any memory profiling to see if there are any leaks?

src/backend/pipeline/cont_adhoc.c
loadzero: select xxx(); functions execute inside portals.

This code does a lot of stuff that is incompatible with that. It is easier to intercept before we get to that point, rather than get there and undo a bunch of stuff.

derekjn: @usmanm this is what we were originally going to do, but it turns out to be pretty complicated to do this cleanly within an actual function call. In particular, function calls don't have access to the client connection to write tuples to. We could probably find a way to work around that, but I don't think that it would be any cleaner than doing this signature-based intercept.

usmanm: Makes sense, might be worth putting a comment describing why we intercept the function early.

src/backend/pipeline/cont_scheduler.c
loadzero: AdhocMgrPeriodicDrain is probably a better name. It is currently possible to insert tuples into the adhoc tuple buffer that won't be read (because the adhoc process has finished). The Drain function scans the adhoc tuple buffer for these kinds of tuples and removes them.

If it is not called, eventually the buffer will fill up and deadlock.

A better design or better synchronization between reader / writers could obviate this.

usmanm: We're moving to a per-process TupleBuffer model, so that should probably alleviate the need for this.

src/backend/pipeline/cont_scheduler.c
loadzero: The function needs to access and modify the pipeline_query, which needs to happen per database.

src/backend/pipeline/cont_adhoc.c
loadzero: Do you mean the call to set_reader? Because the planstate is being reinitialized by ExecInitNode on each run, so set_reader does need to be called.

usmanm: Oh, you're right, my bad! Previously, we used to call `ExecInitNode` and `ExecEndNode` only once.

src/backend/pipeline/cont_adhoc.c
loadzero: TupleBufferBatchReaderRewind resets the cq_id on the reader, so we need to set it again here.

usmanm: False alarm again, hah.

loadzero:
@usman

PR #1225, opened 2015-10-17
Add sleep timeout as a fallback to prevent deadlocks

derekjn:


src/backend/pipeline/tuplebuf.c
```diff
@@ -614,7 +614,7 @@ TupleBufferTryWait(TupleBufferReader *reader)
 	SpinLockRelease(&buf->mutex);
 
 	if (should_wait)
-		WaitLatch(reader->proc->latch, WL_LATCH_SET | WL_POSTMASTER_DEATH, 0);
+		WaitLatch(reader->proc->latch, WL_LATCH_SET | WL_POSTMASTER_DEATH, 1);
```
usmanm: So the problem was that the combiners weren't waking up to read from the CombinerTupleBuffer?

derekjn: Correct.

src/backend/pipeline/cont_scheduler.c
```diff
@@ -673,6 +673,10 @@ ContQuerySchedulerMain(int argc, char *argv[])
 	ContQuerySchedulerShmem->scheduler_pid = MyProcPid;
 
 	dbs = get_database_list();
+	if (list_length(dbs) * TOTAL_SLOTS > max_worker_processes)
```
usmanm: We should probably throw an error and not start any continuous query procs. Seeing the state where one db gets all the workers and the other one doesn't seems like something we shouldn't do.

src/backend/pipeline/tuplebuf.c
usmanm: This fix might not be too great then because this means that our background processes will never go to sleep.

derekjn: We'll they'll sleep for at least 1ms, so when they're not busy they'll spend the majority of their time asleep right?

usmanm: I guess, it's basically going be like having `list_length(db) * TOTAL_SLOTS* processes in a loop like:
```
while (TupleBufferIsEmpty(buf))
  pg_usleep(1000);
```

derekjn: I just ran that loop with 32 procs and the cpu usage is negligible. How about we increase to 10ms and only wait on combiners?

usmanm: Oh, in that case, keep it at 1ms. I forget how fast computers are.

derekjn: I'll leave it at 10 and only on combiners, just to be sure. It's a fallback for a case that shouldn't be happening anyways...

PR #1231, opened 2015-10-25
Add ncurses dependency

usmanm:


configure
```diff
@@ -584,10 +584,10 @@ MAKEFLAGS=
 PACKAGE_NAME='PipelineDB'
 PACKAGE_TARNAME='pipelinedb'
 PACKAGE_VERSION='9.4.4'
-PACKAGE_STRING='PostgreSQL 9.4.4'
+PACKAGE_STRING='PipelineDB 9.4.4'
```
derekjn: Changing these constants is actually pretty sensitive and will cause issues with our binary builds. If you want to keep this change in, we'll want to run all the buildbot builds against this change first.

configure
```diff
@@ -10963,7 +10983,7 @@ else
     We can't simply define LARGE_OFF_T to be 9223372036854775807,
     since some C++ compilers masquerading as C compilers
     incorrectly reject 9223372036854775807.  */
-#define LARGE_OFF_T (((off_t) 1 << 62) - 1 + ((off_t) 1 << 62))
+#define LARGE_OFF_T ((((off_t) 1 << 31) << 31) - 1 + (((off_t) 1 << 31) << 31))
```
derekjn: Let's leave all these how they were; these constants don't need to change.

usmanm: The constants are the same. It's just doing `(1 << 31) << 31` instead of `1 << 62`. `autoconf` prolly changed this to fix some overflow error.

derekjn: Touché

configure
usmanm: Let me try running buildbot on this branch. I just wanted to minimize manual changes we're making to the generated `configure` script.

Issue #1213, opened 2015-10-13
Strange behavior in Node client


derekjn:
(From @EliSnow, originally part of #1204)

Here's a pared down js file which replicates the issue (let me know if you need more than this):
```js
"use strict";
let pg = require("pg");

let connectString = 'postgres://db_user:password@192.168.99.100/stats';
let delimiter = '\r\n';

pg.connect(connectString, (err, client, done) => {
  if (err) {
    console.error(err);
    return;
  }
  let data = '{"foo":"bar"}\r\n{"baz":"bux"}\r\n';
  client.query(
    {
      name: "junk",
      text: `INSERT INTO tsd_stream (js) SELECT unnest(string_to_array(left($1, -${delimiter.length}), $2)::jsonb[])`,
      values: [data, delimiter]
    },
    (err, response) => {
      done();
      if (err) {
        console.error(err);
        console.log(data);
      }
    });
});
```
This is using the npm library [ps](https://www.npmjs.com/package/pg) so a `npm install ps` will resolve that dependency. The stream `tsd_stream` has a single jsonb column `js`.

I'm actually experiencing two issues:
1) in the version above the error returned is `error: expected 1 prepared parameters but received 2`. 
2) my original version did not have `delimiter` as a parameter rather it was concatenated directly into the query string (since it is hardcoded into the server code, it is "safe"). However that resulted in bizarre parsing errors on each of the continuous views feeding off of `tsd_stream`. To demonstrate this problem if you have one or more continuous views using `tsd_stream` and you replace the above query object with:
```js
{
    name: "junk",
    text: `INSERT INTO tsd_stream (js) SELECT unnest(string_to_array(left($1, -${delimiter.length}), '${delimiter}')::jsonb[])`,
    values: [data]
}
```
In the database logs you will a json parsing error on every continuous view feeding off of `tsd_stream`. 
```
ERROR:  invalid input syntax for type json
DETAIL:  Expected end of input, but found "{".
```
If you change the `INSERT` into a `SELECT`, you will see the json is parsed correctly:
```js
{
    name: "junk",
    text: `SELECT unnest(string_to_array(left($1, -${delimiter.length}), '${delimiter}')::jsonb[])`,
    values: [data]
}
```
Which seems to indicate the error is happening when PipelineDB is parsing the query.

EliSnow:
Let me know if you need help replicating. I'm running in Docker containers and I could put together a few Dockerfiles to demo the issues relatively easily.

[...]

Issue #1170, opened 2015-09-15
stream_insert_subselect failure


usmanm:
http://buildbot.pipelinedb.com:8010/builders/CentOS%206.4%20RPM%20Release/builds/0/steps/shell_7/logs/stdio

Issue #1135, opened 2015-08-26
Don't shortcut stream insertions


usmanm:

derekjn:
This should fix #958 and #1213

Issue #958, opened 2015-06-06
Make prepared inserts support expressions


derekjn:
Currently only the raw parameters are used:

```
# create continuous view v as select s::text from stream;
CREATE CONTINUOUS VIEW
# prepare p as insert into stream (s) values (upper($1));
PREPARE
execute p('derek');
EXECUTE
# select * from v;
   s   
-------
 derek
(1 row)

derekjn:
We should rework the prepared insert path to just use the regular insert path. This should be much simpler now that we evaluate stream `INSERT` expressions/queries in a standard `Portal`.

PR #1235, opened 2015-10-26
1135 stream fdw

derekjn:
Streams are now foreign tables accessed view `stream_fdw`. This allows us to use the standard insert and planner paths for stream insertions and scans, so this is a pretty big code quality improvement. All of the special casing for prepared inserts has been ripped out, so this diff fixes a few issues:

* #1135 
* #1213 
* #958 
* #1170  (probably)


usmanm:
Seems to have some merge issue. Changes from other commits are visible here. Might have to rebase?

src/backend/storage/lmgr/deadlock.c
```diff
@@ -948,6 +948,10 @@ DeadLockReport(void)
 
 	pgstat_report_deadlock();
 
+	elog(LOG, "DEADLOCK %d", getpid());
```
usmanm: Remove.

src/backend/utils/cache/typcache.c
```diff
@@ -1265,8 +1265,6 @@ set_record_type_typemod(int32 typmod, TupleDesc desc)
 void
 reset_record_type_cache(void)
 {
-	Assert(IsContQueryWorkerProcess() || IsContQueryAdhocProcess());
```
usmanm: Maybe add `Assert(IsContQueryProcess());`?

src/backend/utils/error/assert.c
```diff
@@ -53,7 +53,7 @@ ExceptionalCondition(const char *conditionName,
 	/* Usually this shouldn't be needed, but make sure the msg went out */
 	fflush(stderr);
 
-#ifdef SLEEP_ON_ASSERT
+#ifndef SLEEP_ON_ASSERT
```
usmanm: Revert.

src/backend/utils/cache/typcache.c
derekjn: The problem is that it gets called from `EXPLAIN`. That didn't used to be the case because this was called in `EndStreamScanBatch`, which doesn't happen anymore.

src/test/regress/expected/foreign_data.out
```diff
@@ -150,13 +160,14 @@ HINT:  Must be superuser to alter a foreign-data wrapper.
...
- foo        | foreign_data_user | -       | -                        |                   | (b '3', c '4', a '2', d '5') | 
- postgresql | foreign_data_user | -       | postgresql_fdw_validator |                   |                              | 
-(3 rows)
+                                                          List of foreign-data wrappers
+    Name    |       Owner       |      Handler       |        Validator         | Access privileges |         FDW Options          | Description 
+------------+-------------------+--------------------+--------------------------+-------------------+------------------------------+-------------
+ dummy      | foreign_data_user | -                  | -                        |                   |                              | useless
+ foo        | foreign_data_user | -                  | -                        |                   | (b '3', c '4', a '2', d '5') | 
+ postgresql | foreign_data_user | -                  | postgresql_fdw_validator |                   |                              | 
+ stream_fdw | derek             | stream_fdw_handler | -                        |                   |                              |
```
usmanm: Owner `derek` seems like will break things.

src/test/regress/expected/typed_streams.out
```diff
@@ -4,11 +4,10 @@ SET search_path TO test_typed_streams,public;
 CREATE STREAM s0 (x integer, y integer);
 DROP TABLE s0;
 ERROR:  "s0" is not a table
-HINT:  Use DROP STREAM to remove a stream.
+HINT:  Use DROP FOREIGN TABLE to remove a foreign table.
```
usmanm: This seems wrong.

src/test/regress/expected/foreign_data.out
derekjn: It's `regression` now.

src/test/regress/pg_regress.c
```diff
@@ -2552,6 +2552,8 @@ regression_main(int argc, char *argv[], init_function ifunc, test_function tfunc
 				break;
 		}
 
+		printf("PGPORT=%s", getenv("PGPORT"));
```
usmanm: Remove.

src/test/regress/expected/typed_streams.out
derekjn: Yea, it kind of sucks but that's all we can do without having a stream relkind:

https://github.com/pipelinedb/pipelinedb/pull/1235/files#diff-2b1247515e9aec2cf5fc7b716eb677c7L256

src/test/regress/expected/typed_streams.out
```diff
@@ -17,12 +16,13 @@ HINT:  Streams can only be read by a continuous view's FROM clause.
...
- x                 | integer
- y                 | integer
- arrival_timestamp | timestamp(0) with time zone
+                   Foreign table "test_typed_streams.l1"
+      Column       |            Type             | Modifiers | FDW Options 
+-------------------+-----------------------------+-----------+-------------
+ x                 | integer                     |           | 
+ y                 | integer                     |           | 
+ arrival_timestamp | timestamp(0) with time zone |           | 
+Server: pipeline_streams
```
usmanm:
```
usmanm=# DROP SERVER pipeline_streams;
DROP SERVER
usmanm=# CREATE STREAM s (x int);
ERROR:  server "pipeline_streams" does not exist
```

src/test/regress/expected/typed_streams.out
usmanm: Playing around with this a bit more, I'm not sure if the quirkiness being introduced merits removing `RELKIND_STREAM`. We probably don't need `RTE_STREAM`, but `RELKIND_STREAM` seems to be the only way we can ensure good usability. But your call. 
```
usmanm=# CREATE STREAM s (x int);
CREATE STREAM
usmanm=# ALTER STREAM s ADD COLUMN y text;
ALTER STREAM
usmanm=# ALTER TABLE s ADD COLUMN z text;
ALTER TABLE
usmanm=# DROP FOREIGN TABLE s;
DROP FOREIGN TABLE
```

derekjn: We should probably talk more about the best balance to strike here. Whatever we end up doing will trade aestheticism for internal code complexity. Adding another relkind essentially means that we have to check for `RELKIND_STREAM` wherever `RELKIND_FOREIGN_TABLE` is referenced.

PR #1242, opened 2015-10-28
Add RELKIND_STREAM

usmanm:


derekjn:
Nice man, looks good! Putting this back was a good call.

Issue #1210, opened 2015-10-11
first and last aggregates


dev1ant:
Something like the following:
https://wiki.postgresql.org/wiki/First/last_%28aggregate%29
https://github.com/wulczer/first_last_agg.git

usmanm:
This is tricky because the ordering of tuples is undefined in streams.

dev1ant:
Doesn't arrival_timestamp provide ordering?

usmanm:
Yea, if that's what you had in mind, then we can support these aggregates.

dev1ant:
Yep, thats it. Thanks in advance.

usmanm:
No worries. Will get it in sometime today.

derekjn:
I'm thinking a signature such as: `first(ts timestamp, value any)` and `first(value any)` for implicit arrival ordering.

usmanm:
We could even have something like `first(cmp_value any, value any)` which would behave the same way as `first(value any) ORDER by cmp_value`.

derekjn:
@dev1ant the way we ended up implementing this is with keyed min/max aggregates. So for first and last arrival, you can now do:

    ...
    SELECT keyed_min(arrival_timestamp, value), keyed_max(arrival_timestamp, value)
    FROM stream
    ...

Obviously this lets you use any field you want to provide ordering. Thanks for the useful feature request!

dev1ant:
Cool, thanks. Will try it soon.

PR #1243, opened 2015-10-28
#1210 Add keyed min/max aggregate

derekjn:


src/test/regress/expected/keyed_min_max.out
```diff
@@ -0,0 +1,158 @@
+SELECT keyed_min(x, -x) FROM generate_series(1, 1000) AS x;
+ keyed_min 
+-----------
+        -1
+(1 row)
+
+SELECT keyed_min(-x::float, x::int8) FROM generate_series(1, 1000) AS x;
+ keyed_min 
+-----------
+         1
```
usmanm: This seems incorrect? Shouldn't the min be `1000`?

derekjn: Yea, I'm making a couple more changes, I'll reopen in a sec.

PR #1244, opened 2015-10-29
#1210 Add keyed min/max aggregate

derekjn:


src/include/catalog/pg_type.h
```diff
@@ -668,6 +668,12 @@ DESCR("filtered space saving");
 DATA(insert OID = 5042 ( _fss	PGNSP PGUID -1 f b A f t \054 0  5041 0 array_in	array_out array_recv array_send - - array_typanalyze i x f 0 -1 0 0 _null_ _null_ _null_ ));
 DESCR("filtered space saving array");
 
+/* key-value pair */
```
usmanm: Do we really need this extra type? I think we can just store the aggregate state as `bytea` and it should work fine.

derekjn: I suppose not, I'll remove it.

usmanm:
Looks good!

PR #1247, opened 2015-10-30
Add fallback for <stdatomic.h>

usmanm:


loadzero:
Looks good to me. Compiles OK in both normal and fallback mode.

PR #1258, opened 2015-11-04
Rejigger + clean up for adhoc queries

usmanm:


loadzero:
The diff looks fine. If it passes the tests, it should be good to go. 

I will take a closer look at the ResourceOwner / transaction changes, as I ran into a few non-obvious corner cases with that stuff.

PR #1262, opened 2015-11-05
Use hash of query name instead of query id

derekjn:


usmanm:
Seems to be breaking tests

usmanm:
Looks good!

PR #1275, opened 2015-11-19
1201 exact distincts

derekjn:


src/test/regress/expected/cont_set_agg.out
```diff
@@ -0,0 +1,65 @@
+CREATE CONTINUOUS VIEW test_set_agg0 AS SELECT set_cardinality(set_agg(x::integer)), exact_count_distinct(x) FROM test_set_agg_stream;
```
usmanm: Can we explicitly check if this works when hash values collide?

src/backend/utils/adt/setfuncs.c
```diff
@@ -0,0 +1,259 @@
...
+#include "utils/datum.h"
+#include "utils/hsearch.h"
+#include "utils/setfuncs.h"
+#include "utils/typcache.h"
+
+#define MURMUR_SEED 0x02cffb4c45ee1fb8L
+
+typedef struct SetKey
+{
+	uint32 k;
```
usmanm: Maybe rename `k` and `d` to `hash` and `value`.

usmanm:
Looks good!

src/test/regress/expected/cont_set_agg.out
derekjn: I checked manually by hard-coding a collision, might take a bit to find a murmur collision to use but I can do it.

PR #1281, opened 2015-11-24
Database level control for activating/deactivating continuous queries

usmanm:
Also adds a DB level meta json field which we can use with extensions


src/backend/catalog/pipeline_database.c
```diff
@@ -0,0 +1,78 @@
...
+	values[Anum_pipeline_database_cq_enabled - 1] = BoolGetDatum(continuous_queries_enabled);
+	nulls[Anum_pipeline_database_meta - 1] = true;
+
+	tup = heap_form_tuple(pipeline_database->rd_att, values, nulls);
+	simple_heap_insert(pipeline_database, tup);
+	CatalogUpdateIndexes(pipeline_database, tup);
+
+	CommandCounterIncrement();
+
+	heap_close(pipeline_database, NoLock);
```
loadzero: Is this meant to be NoLock?

usmanm: Yes, it will automatically be released when the transaction commits.

src/backend/commands/pipelinecmds.c
```diff
@@ -706,3 +707,65 @@ ExecExplainContViewStmt(ExplainContViewStmt *stmt, const char *queryString,
...
+
+	if (row->cq_enabled != value)
+	{
+		bool replace[Natts_pipeline_database];
+		bool nulls[Natts_pipeline_database];
+		Datum values[Natts_pipeline_database];
+		HeapTuple new;
+
+		MemSet(replace, 0 , sizeof(replace));
+		MemSet(nulls, 0 , sizeof(replace));
```
loadzero: should be
```
sizeof(nulls)
```

src/backend/catalog/pipeline_database.c
```diff
@@ -0,0 +1,78 @@
...
+ */
+void
+CreatePipelineDatabaseCatalogEntry(Oid dbid)
+{
+	Relation pipeline_database = heap_open(PipelineDatabaseRelationId, RowExclusiveLock);
+	Datum values[Natts_pipeline_database];
+	bool nulls[Natts_pipeline_database];
+	HeapTuple tup;
+
+	MemSet(nulls, 0, sizeof(nulls));
```
loadzero: should values be zeroed out here too?

usmanm: They're set below for all non-null fields, so there's no need to zero them out.

PR #1301, opened 2015-12-04
Add primary keys to matrels

usmanm:


derekjn:
Looks good. So the only measurable perf impact this should have is on inserts, and even then it's probably negligible. Is that a fair assumption?

usmanm:
Yes, updates should be a little slower since we're writing to an additional index. I haven't done any perf testing, but hopefully the impact should be negligible.

PR #1302, opened 2015-12-04
Alias array_agg(DISTINCT) with set_agg

usmanm:


derekjn:
Looks good!

Added https://github.com/pipelinedb/pipelinedb/issues/1303

Issue #1268, opened 2015-11-12
Better IO throttling of stats collector


usmanm:
Seems to be causing high IO and sometimes is not responsive:
```
LOG:  using stale statistics instead of current ones because stats collector is not responding
LOG:  using stale statistics instead of current ones because stats collector is not responding
LOG:  using stale statistics instead of current ones because stats collector is not responding
LOG:  using stale statistics instead of current ones because stats collector is not responding
LOG:  using stale statistics instead of current ones because stats collector is not responding
LOG:  using stale statistics instead of current ones because stats collector is not responding
STATEMENT:  SELECT COUNT(*) FROM pipeline_proc_stats WHERE type = 'worker'
```

derekjn:
@usmanm did you have a query or load test that can easily repro this locally? We were forcing a stats flush at the beginning of every batch, so 0.8.5 probably doesn't have this issue because of `continuous_query_commit_interval`.

I just disabled the force because it's not necessary anyways.

usmanm:
I think it was when I was testing on the 32 core machine. We also force stream stats at the end of every commit, which might be an issue.

derekjn:
I guess the best we can do is only force a flush when a backend is shutting down:

https://github.com/pipelinedb/pipelinedb/blob/master/src/backend/postmaster/pgstat.c#L2634

That creates a race in a small test but PG just waits in the `stats` test because it has the same issue.

usmanm:
That sounds like the best thing to do.

On Fri, Dec 11, 2015 at 10:22 AM, Derek Nelson <notifications@github.com>
wrote:

> I guess the best we can do is only force a flush when a backend is
> shutting down:
>
>
> https://github.com/pipelinedb/pipelinedb/blob/master/src/backend/postmaster/pgstat.c#L2634
>
> That creates a race in a small test but PG just waits in the stats test
> because it has the same issue.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pipelinedb/pipelinedb/issues/1268#issuecomment-164008950>
> .
>



-- 
Usman

PR #1309, opened 2015-12-11
#1268 Throttle stream stats collection

derekjn:


usmanm:
Looks good--not sure why the test is still failing.

derekjn:
I just ripped the test out, it's not worth the trouble. It has to do with how `SQLAlchemy` pools connections and sometimes the backend isn't terminated

PR #1323, opened 2015-12-21
Fix offset error in pipeline_kafka \n check.

cleaton:
buf.data[buf.len] !=  '\n' is always true because appendBinaryStringInfo null terminates message payload.


derekjn:
Thanks @cleaton!

Merged into `master`.

PR #1329, opened 2015-12-22
Fixes #1328: Log appropriate notice when attempting to DROP STREAM IF…

### LINKED ISSUE #1328: DROP STREAM IF EXISTS should log a notice if the stream doesn't exist
Currently it logs this:

    ERROR:  unexpected object type (36)

derekjn:
… EXISTS


usmanm:
Looks good!

Issue #1206, opened 2015-10-07
Allow indexes to be created directly on continuous views


EliSnow:
Currently if someone wants to index their continuous view they have to add an index on `<continuous view name>_mrel0`. This has two downsides: 1) It's not intuitive and 2) it is impossible to index a column represented by a transition state.

EliSnow:
An alternative fix for point 2 above would be to expose a public function which would take a transition state from the `_mrel` table and return the consumable output.

usmanm:
Hey @EliSnow, you can use the `finalize` function to transform the transition state into the actual value of the aggregate. But you wouldn't be able to `combine` the finalized.

usmanm:
Also, is this urgent for you guys? I can mark it as high priority and get it out soon.

EliSnow:
No, not urgent at all. I will look at the `finalize` function. I guess I missed that one.

usmanm:
I think we might just have not documented it--oops.

PR #1330, opened 2015-12-22
#1206 Support the creation of indices directly on continuous views

derekjn:


src/backend/pipeline/cont_plan.c
```diff
@@ -323,3 +325,93 @@ UnsetEStateSnapshot(EState *estate)
...
+/*
+ * ProcessUtilityOnContView
+ *
+ * Hook to intercept relevant utility queries run on continuous views
+ */
+void
+ProcessUtilityOnContView(Node *parsetree, const char *sql, ProcessUtilityContext context,
+													  ParamListInfo params, DestReceiver *dest, char *tag)
+{
+	if (IsA(parsetree, IndexStmt))
```
usmanm: We should also get rid of `pipeline_rewrite` and intercept `VacuumStmt` here.

src/test/regress/sql/cont_index.sql
```diff
@@ -0,0 +1,29 @@
+/*
```
usmanm: Uncomment?

usmanm:
:+1:

PR #1332, opened 2015-12-23
Add qual support to tuplestore scan node.

loadzero:
This is so where clause filtering works when running from a tuple store.


usmanm:
Looks good!

usmanm:
Something weird going on with that test. Wondering if this broke something in `sw_vacuum.c`.

loadzero:
It turns out that I don't need this. Closing for now.

PR #1336, opened 2016-01-14
Update docker env's continuous_query_ipc_shared_mem unit from mb to MB.

myechuri:
Starting PipelineDB in a container using configuration file https://github.com/pipelinedb/pipelinedb/blob/master/pkg/docker/scripts/conf/pipelinedb.conf fails with the following error:

_______
ALTER ROLE
LOG:  received smart shutdown request
LOG:  autovacuum launcher shutting down
LOG:  continuous query scheduler shutting down
LOG:  shutting down
waiting for server to shut down....LOG:  database system is shut down

 done
server stopped
LOG:  invalid value for parameter "continuous_query_ipc_shared_mem": "8mb"
HINT:  Valid units for this parameter are "kB", "MB", "GB", and "TB".
FATAL:  configuration file "/mnt/pipelinedb/data/pipelinedb.conf" contains errors
_______

This PR attempts to fix the unit for continuous_query_ipc_shared_mem in pipelinedb.conf.

Tested the fix, db container came up fine:
_______
LOG:  database system is ready to accept connections
LOG:  continuous query scheduler started
LOG:  autovacuum launcher started
_______

### DIFF
--- a/pkg/docker/scripts/conf/pipelinedb.conf
+++ b/pkg/docker/scripts/conf/pipelinedb.conf
@@ -85,7 +85,7 @@ continuous_query_combiner_work_mem = 16MB
 
 # amount of shared memory reserved by each continuous query process
 # for IPC
-continuous_query_ipc_shared_mem = 8mb
+continuous_query_ipc_shared_mem = 8MB
 
 # the default step factor for sliding window continuous queries (as a percentage
 # of the total window size)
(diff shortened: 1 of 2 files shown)


usmanm:
Copied your change and ported it over to the `pipelinedb.conf.sample` file as well. Thanks @myechuri!

myechuri:
Great! Thanks for the quick action, @usmanm !

PR #1340, opened 2016-01-19
add activate and deactivate for tab completion

guanqun:


usmanm:
Looks good, thanks @guanqun!

PR #1346, opened 2016-01-24
add instruction how to test pipeline

guanqun:
Make it clear that pipeline uses ``make check`` instead of ``make test`` to do unit testing.


usmanm:
Thanks @guanqun!

PR #1353, opened 2016-02-02
Add non-blocking version of logical_read_local_xlog_page

loadzero:
Add support for using logical decoding in a non blocking fashion.


usmanm:
Looks good!

PR #1361, opened 2016-02-09
SELECT ... FOR UPDATE when looking up groups

usmanm:


src/backend/executor/nodePhysicalGroupLookup.c
```diff
@@ -45,24 +48,89 @@ ExecInitPhysicalGroupLookup(PhysicalGroupLookup *node, EState *estate, int eflag
...
+	EState *estate = outer->js.ps.state;
+	Buffer buffer;
+	HeapUpdateFailureData hufd;
+	HeapTuple tup;
+	Relation rel;
+	TupleTableSlot *slot = NULL;
+	BitmapHeapScanState *scan;
 
+	/* Get next tuple from subplan, if any. */
+lnext:
```
loadzero: Thats an interesting use of goto. What did the non-goto version look like?

derekjn:
Looks good to me :+1:

src/backend/executor/nodePhysicalGroupLookup.c
usmanm: There wasn't need for any retry before since we didn't do any locking of tuples.

loadzero: Sorry, I wasn't being clear. I meant that I find backwards jumping goto harder to read and more error-prone compared to a loop/break/continue construct.

I was wondering whether there was a clarity reason for doing it this way.

usmanm: That's how Postgres does it in `nodeLockRows.c`. This logic is pretty similar to what goes inside that.

loadzero: Gotcha. Makes sense now.

PR #1383, opened 2016-02-20
Reduce logging noise

derekjn:
I just demoted the log level to `DEBUG1`, which seems like the correct level anyways...


usmanm:
👍🏿

PR #1393, opened 2016-02-26
Continuous transforms

usmanm:


derekjn:
Am I doing something dumb here?

```
create continuous view v as select x::integer from even_stream;
create continuous transform tv as
    select x::integer from stream where x % 2 = 0
    then execute procedure pipeline_stream_insert('even_stream');

insert into stream (x) values (2);
select * from v;
 x 
---
      <-- null
(1 row)
```

usmanm:
synchronous_stream_insert doesn't work right now. Could that be an issue?

derekjn:
I don't think so. If I insert another row, there are two null rows in the CV, so it seems to be reading something.

usmanm:
I'll look into it in a little bit when I get home.

On Saturday, February 27, 2016, Derek Nelson <notifications@github.com>
wrote:

> I don't think so. If I insert another row, there are two null rows in the
> CV, so it seems to be reading something.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pipelinedb/pipelinedb/pull/1393#issuecomment-189769037>
> .
>


-- 
Usman

usmanm:
Should be fixed now.

derekjn:
This is awesome man. The code looks good as far as I can tell, I'm going to continue spending some time functionally testing it though. I'll also add a few tests to this branch.

derekjn:
`catversion.h:CATALOG_VERSION_NO` needs to be bumped.

usmanm:
Also need to add support for `pg_dump` and `pg_restore`.

derekjn:
I'm also getting `NULL` rows in continuous view reading from a continuous xform doing an STJ:

```
create continuous transform x0 as
    select t.s as s from raw_stream s join t on t.x = s.x::integer
    then execute procedure pipeline_stream_insert('stream');
create continuous view v0 as select s::text from stream;
create continuous view v1 as select s::text from stream;

insert into stream (x) values (0);
select * from v0;
 s 
---
 
(1 row)
```

src/backend/pipeline/transformReceiver.c
```diff
@@ -0,0 +1,242 @@
...
+	MemoryContextSwitchTo(old);
+}
+
+static void
+transform_destroy(DestReceiver *self)
+{
+	TransformState *t = (TransformState *) self;
+	pfree(t);
+
+	elog(LOG, "destroy");
```
derekjn: Remove

derekjn:
re: my previous comment, I forgot to mention that before doing the stream insertion I did:

```
create table t (x integer, s text);
insert into t (x, s) values (0, 'zero');
insert into t (x, s) values (1, 'one');
```

Also, it appears that `transform_receive` isn't even being called after the stream insertion. It doesn't appear that `init_query_state` is being called for the xform either.

derekjn:
Disregard previous comments about null rows, I was doing something dumb.

PR #1406, opened 2016-03-10
Lock tuples in ascending order when looking them up

usmanm:


derekjn:
Looks good :+1:

usmanm:
Turns out this actually might not fix the problem, we'll just have to stick to an `ExclusiveLock`: http://stackoverflow.com/questions/10245560/deadlocks-in-postgresql-when-running-update

PR #1434, opened 2016-04-05
1433 trigger process cleanup

loadzero:
There is still a race condition lurking here I suspect, but this should fix the main problem. Will need to revisit this.

### DIFF
--- a/src/test/py/test_crash_recovery.py
+++ b/src/test/py/test_crash_recovery.py
@@ -6,6 +6,7 @@ import random
 import signal
 import threading
 import time
+import sys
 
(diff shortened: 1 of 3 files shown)


usmanm:
Looks good!

src/test/py/test_crash_recovery.py
```diff
@@ -218,6 +218,8 @@ def test_activate_deactivate(pipeline, clean_db):
   cur.close()
   conn.close()
 
+  time.sleep(2)
```
usmanm: Do we need this sleep?

loadzero: Yes, the process deactivation now takes longer. I suspect there was a race beforehand that is now being exposed.

PR #1438, opened 2016-04-05
1432 alert server ports

loadzero:
Clean up alert server port assignment and error handling.


src/backend/utils/misc/guc.c
```diff
@@ -3597,7 +3606,15 @@ static struct config_string ConfigureNamesString[] =
 		NULL,
 		NULL, NULL, NULL,
 	},
-
+	{
+		{"alert_server_address", PGC_POSTMASTER, CUSTOM_OPTIONS,
```
usmanm: Can't we use `listen_addresses`?

loadzero: It's a bit clunky. This is the address that the client needs to use. There are a bunch of ways we might guess it, but we would probably still need to provide a way to override it.

PR #1446, opened 2016-04-07
Abort batch transaction if trigger function fails.

loadzero:
If a trigger function inside a continuous trigger batch to be processed fails, abort the transaction and clean up.

### DIFF
--- a/src/test/regress/sql/cont_trigger.sql
+++ b/src/test/regress/sql/cont_trigger.sql
@@ -1,4 +1,6 @@
-CREATE CONTINUOUS VIEW cont_tg_cv AS SELECT count(*) FROM cont_tg_stream;
+select pg_sleep(2);
+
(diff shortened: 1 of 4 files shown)


src/test/regress/sql/cont_trigger.sql
```diff
@@ -1,4 +1,6 @@
-CREATE CONTINUOUS VIEW cont_tg_cv AS SELECT count(*) FROM cont_tg_stream;
+select pg_sleep(2);
```
usmanm: Why is this needed?

usmanm:
Looks good! If that sleep was added by accident, you can push this after removing it.

src/test/regress/sql/cont_trigger.sql
loadzero: The trigger process isn't started until after that sleep. The sleep was there before, I just moved it to the start. A better solution might be to block on creating triggers until the trigger proc is up.

PR #1454, opened 2016-04-08
1449 triggers finalize

loadzero:
Pass the correct rel into trigger context.
Add a regression test.


usmanm:
Can we also add a test where the condition is checking against a finalized value?

usmanm:
I think we also don't need to open the matrel in `process_batch` and pass it down to the trigger func.

PR #1457, opened 2016-04-11
Check continuous trigger requirements.

loadzero:
Added a function for checking replication slot and wal sender requirements for trigger proc.

It is used in cont_scheduler (start_database_workers), before launching the trigger proc, and logs a warning if requirements aren't met. We should probably change the semantic to not activate the database.

It is checked again inside trigger_main before the main loop, and will error out if requirements aren't met.

### DIFF
--- a/src/backend/pipeline/cont_scheduler.c
+++ b/src/backend/pipeline/cont_scheduler.c
@@ -49,6 +49,9 @@
 #include "utils/timeout.h"
 #include "pipeline/trigger/trigger.h"
 #include "pipeline/trigger/triggerfuncs.h"
+#include "replication/walsender.h"
+#include "replication/walsender_private.h"
+#include "replication/slot.h"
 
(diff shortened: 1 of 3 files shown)


src/backend/pipeline/cont_scheduler.c
```diff
@@ -1416,3 +1427,70 @@ AreContQueriesEnabled(void)
...
+bool
+CheckContinuousTriggerRequirements(int elevel)
+{
+	int i = 0;
+	bool free_wal_sender = false;
+	bool free_repl_slot = false;
+
+	if (!XLogLogicalInfoActive())
+	{
+		elog(elevel, "continuous triggers require wal_level set to wal_level_logical");
```
usmanm: logical, not wal_level_logical

usmanm:
Looks good!

PR #1459, opened 2016-04-12
use PushActiveSnapshot in process_batches

loadzero:
This is required for trigger functions that call user defined functions.


usmanm:
Looks good!

Issue #1426, opened 2016-03-30
Performance & stability regression


usmanm:
The last few changes made to solve deadlocks in continuous transforms seems to have caused some performance regressions. Ensure that #1423 keeps performance at par with what was before.

PR #1468, opened 2016-04-21
1423 broker

usmanm:


EliSnow:
Will this also resolve #1426?

usmanm:
This is mainly for fixing deadlocks in transforms. Once that's solved, I'll work on getting perf levels back to 0.9.0.

PR #1472, opened 2016-04-25
Fix deadlock between drop view and combiner.

loadzero:
GetContPlan grabs a lock on the matrel index, before the lock on the matrel was being taken.
This is in reverse order to dropping, so this causes a deadlock.

Moving the matrel lock above GetContPlan fixes this.


loadzero:
I think this deadlock possibility was always there but has recently been exposed by spurious wakeups in the combiner IPC.

The spurious wakeups are caused by a SIGALRM firing, which makes ipc_queue_wait_non_empty return

usmanm:
Where are the SIGALRM signals coming from? I think I can also add the loop back to wait_non_empty now.

loadzero:
When I looked into it, they were coming from schedule_alarm. I did not get a full backtrace, but it looks like they are used in a bunch of places.

usmanm:
That's kinda lame...

Looks good!

PR #1496, opened 2016-05-06
Average and throughput statistics

usmanm:


derekjn:
Awesome, looks good :+1: but we'll need a `CATVERSION` bump since some function signatures changed.

PR #1501, opened 2016-05-09
fix bgw extra len copy

loadzero:
Add missing memcpy back to bgworker.c

We didn't leave this out on purpose right?


usmanm:
Looks good, can you remove the `32917c84ea80a0b637eb3f2d0141c678173bb514` commit from this diff. I force resetted master to remove that commit.

PR #1509, opened 2016-05-12
Bulk copying in IPC broker

usmanm:
Also now there are two separate queues to write into, one for insert procs and one for worker procs. The broker copies messages from those two into the a queue that the worker reads from. If this tends to be a bottleneck we could have the workers read directly from the insert proc queue and the broker only move messages that are being exchanged between workers.

Also fixed potential memory leak in broker process when locally buffering slots.


src/test/regress/parallel_schedule
```diff
@@ -1,120 +1,4 @@
-# ----------
```
derekjn: Whoops

derekjn:
Awesome, looks good! :+1:

PR #1512, opened 2016-05-13
Bulk copy into worker queues when using COPY or transforms

usmanm:


derekjn:
Nice, looks good :+1:

PR #1534, opened 2016-05-26
If no queries were found, discard messages in process queues

usmanm:


derekjn:
Looks good :+1: , thanks man!

PR #1539, opened 2016-05-27
Rename pipeline-server to pipelinedb

usmanm:


derekjn:
Do you think we should symlink `pipeline-server` -> `pipelinedb` for the next couple of releases?

usmanm:
I think most people use `pipeline-ctl` which should keep on working as expected?

derekjn:
Cool, let's just go with it then :+1:

PR #1546, opened 2016-05-27
Add barrier after head/tail modification

derekjn:
This seems to fix the general stability issues we began seeing after https://github.com/pipelinedb/pipelinedb/commit/a5412f8eb59496e482c20be4dccf9acbaf4cb13b and explains why issues appeared to be dependent on optimization as well as OS.


usmanm:
Looks good!

PR #1551, opened 2016-06-01
Store relid in pipeline_query

usmanm:


derekjn:
Looks good, this seems a lot cleaner :+1:

PR #1579, opened 2016-07-24
1573 changestreams

derekjn:
Adds support for output streams, which are turning out to be pretty awesome. Some notes on how this works:

* If anything is reading from an output stream, the combiner will write the old and new tuple to it when syncing, otherwise there's no additional meaningful overhead
* For sliding window queries, output stream writes are done separately by executing the overlay plan on cached in-window tuples
* Sliding-window queries will tick roughly every `step_size_ms` in order to keep writing out values that depend on wall time
* Out-of-window tuples are GC'd and indicated by the output stream tuple of the form: `(<old>, null)`
* Concurrent combiners only cache and process the SW groups that belong to them, so all of this work is distributed evenly across combiners

Note that this diff doesn't include a function wrapper (e.g. `output_of('cv')`) for the output stream yet, so output streams are just referred to literally (`v_osrel`). Doing this the right way isn't too complicated but actually isn't trivial and does require some thought, and I didn't want to go out of scope with this diff.

As soon as this diff is ready to land, I'm going to rip out all legacy trigger stuff. I just didn't want the diff to be unnecessarily large. I think we should rip out the alert server as well.

Also, pytests are on the way.

### DIFF
--- a/src/backend/catalog/pipeline_query.c
+++ b/src/backend/catalog/pipeline_query.c
@@ -211,7 +211,7 @@ GetPipelineQueryTuple(RangeVar *name)
  * Adds a CV to the `pipeline_query` catalog table.
  */
 Oid
-DefineContinuousView(Oid relid, Query *query, Oid matrelid, Oid csrelid, Oid seqrelid, bool gc, bool adhoc, Oid *pq_id)
+DefineContinuousView(Oid relid, Query *query, Oid matrelid, Oid osrelid, Oid seqrelid, bool gc, bool adhoc, Oid *pq_id)
 {
 	Relation pipeline_query;
 	HeapTuple tup;
@@ -242,7 +242,7 @@ DefineContinuousView(Oid relid, Query *query, Oid matrelid, Oid csrelid, Oid seq
 	values[Anum_pipeline_query_relid - 1] = ObjectIdGetDatum(relid);
 	values[Anum_pipeline_query_active - 1] = BoolGetDatum(continuous_queries_enabled);
 	values[Anum_pipeline_query_query - 1] = CStringGetTextDatum(query_str);
-	values[Anum_pipeline_query_csrelid - 1] = ObjectIdGetDatum(csrelid);
+	values[Anum_pipeline_query_osrelid - 1] = ObjectIdGetDatum(osrelid);
 	values[Anum_pipeline_query_matrelid - 1] = ObjectIdGetDatum(matrelid);
 	values[Anum_pipeline_query_seqrelid - 1] = ObjectIdGetDatum(seqrelid);
 	values[Anum_pipeline_query_gc - 1] = BoolGetDatum(gc);
@@ -553,7 +553,7 @@ GetContQueryForId(Oid id)
 	cq->seqrelid = row->seqrelid;
 	cq->matrelid = row->matrelid;
 	cq->active = row->active;
-	cq->csrel = makeRangeVar(get_namespace_name(get_rel_namespace(row->csrelid)), get_rel_name(row->csrelid), -1);
+	cq->osrel = makeRangeVar(get_namespace_name(get_rel_namespace(row->osrelid)), get_rel_name(row->osrelid), -1);
 
 	if (cq->type == CONT_VIEW)
 	{
--- a/src/backend/catalog/pipeline_query.c
+++ b/src/backend/catalog/pipeline_query.c
@@ -550,7 +550,7 @@ GetContQueryForId(Oid id)
 	cq->type = row->type == PIPELINE_QUERY_TRANSFORM ? CONT_TRANSFORM : CONT_VIEW;
 
 	cq->relid = row->relid;
-	cq->relname = get_rel_name(row->relid);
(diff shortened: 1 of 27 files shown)


derekjn:
Also, after doing some profiling on SW ticking, the overlay plan execution is definitely the obvious bottleneck.

usmanm:
Unfortunately, we won't be able to do it for aggregates which are uncombinable. But other than `count(DISTINCT)`, I think all the other commonly used aggregates are uncombinable.

usmanm:
Instead of writing to any worker queue, I think the output streams should be sharded using the hash as well so that the changes made to a specific group are never reordered? Only insert should write to any worker.

derekjn:
Doesn't sharding over workers by combiner group hash mean (in the worst case) acquiring a lock on a worker for every tuple being written to the output stream? I think it's a good idea to always write to the same worker, but shouldn't we choose it like this?

    worker_queue = worker_queues % combiner_id

usmanm:
Yes, that's probably a better way to do it. We also only need to write to the same worker when writing from a combiner process, otherwise we can keep on round robinning.

derekjn:
Cool, change coming shortly.

src/backend/catalog/pipeline_query.c
```diff
@@ -548,10 +550,11 @@ GetContQueryForId(Oid id)
 	cq->type = row->type == PIPELINE_QUERY_TRANSFORM ? CONT_TRANSFORM : CONT_VIEW;
 
 	cq->relid = row->relid;
-	cq->relname = get_rel_name(row->relid);
+	cq->name = makeRangeVar(get_namespace_name(get_rel_namespace(row->relid)), get_rel_name(row->relid), -1);
 	cq->seqrelid = row->seqrelid;
 	cq->matrelid = row->matrelid;
 	cq->active = row->active;
+	cq->osrel = makeRangeVar(get_namespace_name(get_rel_namespace(row->osrelid)), get_rel_name(row->osrelid), -1);
```
usmanm: Better to use `relid` here and use `heap_open_extended` downstream. `relid` can never change even if the relation is renamed etc.

src/backend/pipeline/cont_combiner.c
```diff
@@ -718,6 +1438,43 @@ combine(ContQueryCombinerState *state)
...
+assign_output_stream_projection(ContQueryCombinerState *state)
+{
+	ListCell *lc;
+	bool needs_proj = false;
+	PlannedStmt *overlay = GetContinuousViewOverlayPlan(state->base.query);
+	EState *estate = CreateExecutorState();
+	ExprContext *context = CreateStandaloneExprContext();
+
+	/* Non-sliding windows not supported */
+	if (GetWindowTimeColumn(state->base.query->name))
```
usmanm: Why not? Non-sliding window queries can have finalize functions as well?

src/backend/pipeline/cqmatrel.c
```diff
@@ -28,6 +28,34 @@
...
+	return resultRelInfo;
+}
+
+/*
+ * CQCSRelClose
+ */
+void
+CQOSRelClose(ResultRelInfo *rinfo)
+{
+	ExecCloseIndices(rinfo);
```
usmanm: Not needed.

src/backend/pipeline/stream_fdw.c
```diff
@@ -532,20 +513,44 @@ BeginStreamModify(ModifyTableState *mtstate, ResultRelInfo *result_info,
...
+		 * We always write to the same worker from a combiner process to prevent
+		 * unnecessary reordering
+		 */
+		if (IsContQueryCombinerProcess())
+		{
+			int idx = MyContQueryProc->group_id % continuous_query_num_workers;
+			sis->worker_queue = get_worker_queue_with_lock(idx, false);
+		}
+		else
+		{
```
usmanm: No need for braces.

src/backend/pipeline/stream_fdw.c
```diff
@@ -532,20 +513,44 @@ BeginStreamModify(ModifyTableState *mtstate, ResultRelInfo *result_info,
...
 	sis->num_batches = 1;
-	sis->desc = is_inferred_stream_relation(stream) ? ExecTypeFromTL(insert_tl, false) : RelationGetDescr(stream);
+
+	if (is_inferred_stream_relation(stream))
+	{
+		Assert(insert_tl);
+		sis->desc = ExecTypeFromTL(insert_tl, false);
+	}
+	else
+	{
```
usmanm: No need for braces.

src/backend/pipeline/cont_combiner.c
derekjn: Whoops, that's not supposed to be there.

src/backend/pipeline/stream_fdw.c
derekjn: Technically the style guide says that braces should be used for either all `if` blocks or none :/

usmanm: Oh, weird. I guess that makes sense.

usmanm:
Looks good!

PR #1619, opened 2016-08-10
stream_insert_level and pipeline_flush

usmanm:


derekjn:
Looks good to me, the only thing I'll bring up is nomenclature. `async` and `sync_commit` seem pretty intuitive but `sync_read` might be slightly confusing. Maybe something like `sync_receive` would be more fitting? Or even `sync_ack`? Either way I think `sync_read` is probably fine.

PR #1630, opened 2016-08-21
Fixes #1473: Fix possible deadlocks between combiner and DROP CONTINU…

### LINKED ISSUE #1473: Fix deadlock between ExecCreateContViewStmt and Drop Continuous view
This deadlock occurs when a ```DROP CASCADE``` from ```./src/test/regress/sql/analyze_cont_view.sql``` conflicts with a ```CREATE CONTINUOUS VIEW``` from one of the other regressions tests occurring in parallel

```
DETAIL:  Process 1110 waits for ExclusiveLock on relation 4242 of database 16384; blocked by process 1113.
    Process 1113 waits for AccessShareLock on relation 16755 of database 16384; blocked by process 1110.
    Process 1110: DROP TABLE tnotice CASCADE;
    Process 1113: CREATE CONTINUOUS VIEW cv_agg_expr AS SELECT k::text, COUNT(*) + SUM(v::int) FROM sanity_stream GROUP BY k;
```

```ExecCreateContViewStmt``` takes a lock on the ```PipelineQueryTable``` and then calls ```streams_to_meta```, which acquires locks on all view matrels (via get_query_def).

This is in reverse order to the ```DROP TABLE CASCADE```, which takes a lock on a matrel, and then the ```PipelineQuery``` table.

The relevant backtrace for ```ExecCreateContViewStmt``` is here:

```
src/backend/nodes/print.c:78
LockRelationOid
src/backend/storage/lmgr/lmgr.c:117
relation_open
src/backend/access/heap/heapam.c:1064
heap_open
src/backend/access/heap/heapam.c:1239
AcquireRewriteLocks
src/backend/rewrite/rewriteHandler.c:178
get_query_def
src/backend/utils/adt/ruleutils.c:4380
deparse_query_def
src/backend/utils/adt/ruleutils.c:461
streams_to_meta
src/backend/catalog/pipeline_stream.c:178
UpdatePipelineStreamCatalog
src/backend/catalog/pipeline_stream.c:491
DefineContinuousView
src/backend/catalog/pipeline_query.c:275 (discriminator 4)
ExecCreateContViewStmt
```

derekjn:
There were several things going on here that made it possible for drop/combiner deadlocks. Basically we were just acquiring too many locks and not releasing them aggressively enough:

* In some places we were closing relations with `heap_close(rel, NoLock)` which is problematic when multiple syncs happen in the same transaction
* `select_existing_groups` already has a read lock on the matrel, but the `SELECT` query that runs also acquired a read lock. This one wasn't released until commit, so now we just have a  `EXEC_NO_MATREL_LOCKING` flag that is analagous to `EXEC_NO_STREAM_LOCKING` to prevent multiple locks from being acquired.
* The order in which we acquired osrel and matrel locks was incongruent with the order in which they're ultimately dropped.

### DIFF
--- a/src/backend/executor/execUtils.c
+++ b/src/backend/executor/execUtils.c
@@ -827,6 +827,14 @@ ExecOpenScanRelation(EState *estate, Index scanrelid, int eflags)
 	if ((eflags & EXEC_NO_STREAM_LOCKING) && is_stream_relid(reloid))
 		lockmode = NoLock;
 
+	/*
+	 * Similarly to the above, combiners manage matrel locks themselves and
+	 * so the relation is already locked. Acquiring an extraneous lock here
+	 * can then ultimately cause deadlocks when CVs are being dropped.
+	 */
+	if ((eflags & EXEC_NO_MATREL_LOCKING))
+		lockmode = NoLock;
+
 	rel = heap_open(reloid, lockmode);
 
 	/*
--- a/src/backend/pipeline/cont_combiner.c
+++ b/src/backend/pipeline/cont_combiner.c
@@ -459,7 +459,7 @@ select_existing_groups(ContQueryCombinerState *state)
 	dest = CreateDestReceiver(DestTupleTable);
 	SetTupleTableDestReceiverParams(dest, existing, existing->tablecxt, true);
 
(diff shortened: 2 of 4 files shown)


usmanm:
Looks good!

PR #1656, opened 2016-10-27
Fixes #1655: Don't sync SW matrels if nothing is reading their output…

### LINKED ISSUE #1655: Don't sync SW matrels if nothing is reading their output stream

derekjn:
… stream


usmanm:
Can't we just move `sync_sw_matrel_groups(state, matrel);` from L920 to L953?

derekjn:
Actually yea, that should work.

PR #1674, opened 2016-11-04
1662 ttl

derekjn:
Adds support for TTLs on non-SW CVs. How it works

* TTL and TTL column stored in `pipeline_query`
* Vacuuming logic made slightly more generic to pull TTL expression if there is one
* Otherwise uses exact same logic as previous SW vacuuming
* `max_age` renamed to `sw` for better distinction

### DIFF
--- a/src/backend/catalog/pipeline_query.c
+++ b/src/backend/catalog/pipeline_query.c
@@ -211,7 +211,8 @@ GetPipelineQueryTuple(RangeVar *name)
  * Adds a CV to the `pipeline_query` catalog table.
  */
 Oid
-DefineContinuousView(Oid relid, Query *query, Oid matrelid, Oid seqrelid, int sw_attno, bool adhoc, Oid *pq_id)
+DefineContinuousView(Oid relid, Query *query, Oid matrelid, Oid seqrelid, int ttl,
+		AttrNumber ttl_attno, AttrNumber sw_attno, bool adhoc, Oid *pq_id)
 {
 	Relation pipeline_query;
 	HeapTuple tup;
@@ -244,7 +245,9 @@ DefineContinuousView(Oid relid, Query *query, Oid matrelid, Oid seqrelid, int sw
 	values[Anum_pipeline_query_query - 1] = CStringGetTextDatum(query_str);
 	values[Anum_pipeline_query_matrelid - 1] = ObjectIdGetDatum(matrelid);
 	values[Anum_pipeline_query_seqrelid - 1] = ObjectIdGetDatum(seqrelid);
-	values[Anum_pipeline_query_sw_attno - 1] = ObjectIdGetDatum(sw_attno);
+	values[Anum_pipeline_query_sw_attno - 1] = Int16GetDatum(sw_attno);
+	values[Anum_pipeline_query_ttl - 1] = Int32GetDatum(ttl);
+	values[Anum_pipeline_query_ttl_attno - 1] = Int16GetDatum(ttl_attno);
 	values[Anum_pipeline_query_adhoc - 1] = BoolGetDatum(adhoc);
 
 	if (AttributeNumberIsValid(sw_attno))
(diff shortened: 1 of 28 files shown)


src/backend/catalog/pipeline_query.c
```diff
@@ -561,6 +564,25 @@ IsSWContView(RangeVar *name)
...
+bool
+IsTTLContView(RangeVar *name)
+{
+	bool ttl = false;
+	HeapTuple tuple = GetPipelineQueryTuple(name);
+
+	if (HeapTupleIsValid(tuple))
+	{
+		Form_pipeline_query row = (Form_pipeline_query) GETSTRUCT(tuple);
+		ttl = AttributeNumberIsValid(row->sw_attno) || AttributeNumberIsValid(row->ttl_attno);
```
usmanm: For SW views, we should set the `ttl_attno` to the same as `sw_attno` an change this function to just check that attribute. It seems a little messy otherwise.

src/backend/catalog/pipeline_query.c
```diff
@@ -916,6 +938,8 @@ DefineContinuousTransform(Oid relid, Query *query, Oid typoid, Oid osrelid, Oid
 	values[Anum_pipeline_query_seqrelid - 1] = ObjectIdGetDatum(InvalidOid);
 	values[Anum_pipeline_query_sw_attno - 1] = ObjectIdGetDatum(InvalidOid);
 	values[Anum_pipeline_query_adhoc - 1] = BoolGetDatum(adhoc);
+	values[Anum_pipeline_query_ttl - 1] = Int32GetDatum(0);
+	values[Anum_pipeline_query_ttl_attno - 1] = Int16GetDatum(0);
```
usmanm: `InvalidAttributeNumber` instead of `0`.

src/backend/catalog/pipeline_query.c
```diff
@@ -916,6 +938,8 @@ DefineContinuousTransform(Oid relid, Query *query, Oid typoid, Oid osrelid, Oid
 	values[Anum_pipeline_query_seqrelid - 1] = ObjectIdGetDatum(InvalidOid);
 	values[Anum_pipeline_query_sw_attno - 1] = ObjectIdGetDatum(InvalidOid);
```
usmanm: We should change this to `int16` as well and use `InvalidAttributeNumber` instead.

src/backend/commands/pipelinecmds.c
```diff
@@ -517,6 +518,86 @@ check_relation_already_exists(RangeVar *rv)
...
+
+	if (opt_ttl_col && !opt_ttl)
+		elog(ERROR, "ttl must be specified in conjunction with ttl_column");
+
+	Assert(opt_ttl);
+	Assert(opt_ttl_col);
+
+	*options = list_delete(*options, opt_ttl);
+	*options = list_delete(*options, opt_ttl_col);
+
```
usmanm: Remove new line.

src/backend/catalog/pipeline_query.c
usmanm: Actually, I think we shouldn't even have `sw_attno` now. It should just be a boolean column called `sw`. SW views should simply be TTL views + filtering/combining on read path.

src/backend/pipeline/ttl_vacuum.c
```diff
@@ -200,15 +210,20 @@ NumSWExpiredTuples(Oid relid)
 	resetStringInfo(&sql);
 
 	stmt = (SelectStmt *) linitial(parsetree_list);
-	stmt->whereClause = get_sw_vacuum_expr(cvname);
 
+	if (IsSWContView(cvname))
```
usmanm: We shouldn't need this as well. We should just copy the SW interval to the `ttl` column of the entry in `pipeline_query`.

usmanm:
Everything looks good except that I think we can make the SW stuff cleaner by treating them as TTL + filtering/combining on read path. `pipeline_query` should only store a boolean in the `sw` column, no need for `sw_attno`. The SW expression should be deconstructed and `ttl` and `ttl_attno` should be populated from it. Re-constructing the SW expression also becomes much simpler now which we can do by simply looking up the query in `pipeline_query` rather than having to re-parse the query of storing as extra attribute on the `Query` struct.

derekjn:
I had the same idea but wanted to keep this as surgical as possible and I started to touch too many areas, but it's the right way to do this so I'll rework it a little and unify TTL and SWs more.

derekjn:
So I just removed `sw_attno` because we can reliably indicate SWs with a non-zero `step_factor`.

usmanm:
Are you talking about the `step_factor` in the `Query` struct?

derekjn:
No, in `pipeline_query`.

Issue #1675, opened 2016-11-04
Rip out adhoc metadata


derekjn:
And cleanup logic, etc. It's all dead code.

PR #1676, opened 2016-11-04
#1675: Remove adhoc cruft

usmanm:


derekjn:
Can you please bump the catalog version before committing this to account for the TTL changes as well as this diff's changes?

PR #1687, opened 2016-11-05
Release matrel lock before combining

derekjn:
This seems to have started happening more often after the TTL branch landed, which I'm not totally clear on yet. But basically we were holding a matrel lock while calling `combine`, which acquires the necessary locks on its own.

### DIFF
--- a/src/backend/pipeline/cont_combiner.c
+++ b/src/backend/pipeline/cont_combiner.c
@@ -1982,6 +1982,8 @@ pipeline_combine_table(PG_FUNCTION_ARGS)
 		elog(ERROR, "schema of \"%s\" does not match the schema of \"%s\"",
 				text_to_cstring(relname), quote_qualified_identifier(cv->matrel->schemaname, cv->matrel->relname));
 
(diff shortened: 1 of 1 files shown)


usmanm:
We can't release the lock early. I tried with a `RowExclusiveLock` but that doesn't work because the order in which we acquire the tuple locks might conflict with the combiner. So we need to keep it for the duration of the combine operation. We might have to be a little more surgical here.

derekjn:
Yea, I need to do some more investigating here anyways. After doing some bisecting + running tests locally when new failures showed up on this branch, it looks like 310bce430db9a3a9e67e638c4e5b2a3c8319b9fa might have introduced some instability. In particular it looks like that's when these started happening:

```
WARNING:  timed out waiting for continuous query process "worker0 [pipeline]" to reach state 0
WARNING:  timed out waiting for continuous query process "worker1 [pipeline]" to reach state 0
WARNING:  timed out waiting for continuous query process "combiner0 [pipeline]" to reach state 0
```

I'll investigate and push any fixes on this branch.

PR #1698, opened 2016-11-17
#1672 Skip transforms that don't have a trigger function or output st…

derekjn:
…ream readers


src/backend/pipeline/cont_worker.c
```diff
@@ -314,18 +315,33 @@ ContinuousQueryWorkerMain(void)
...
-
-				ExecutePlan((EState *) estate, state->query_desc->planstate, state->query_desc->operation,
-						true, 0, ForwardScanDirection, state->dest);
-
-				/* free up any resources used by this plan before committing */
-				end_plan(state->query_desc);
-
-				/* flush tuples to combiners or transform out functions */
-				flush_tuples(state);
+				if (state->base.query->type == CONT_TRANSFORM)
```
usmanm: Maybe move to a separate function like `should_exec_query`

PR #1720, opened 2016-12-09
Hll union

derekjn:
This adds an `HLL_UNION` HLL representation that enables optimization for batch `combines`. e.g.,

```sql
select combine(uniques) from v0;
```
```sql
create continuous view v0 as
  select x, count(*), count(distinct y) as uniques from s group by x;
```

It works by loading the union's registers into a `uint8` array rather than performing bit packing/unpacking, when taking the max of registers, which ends up costing a lot of CPU. 

For `select combine(uniques) from v0` run on 100,000 dense HLLs in the above CV, this optimization results in a ~2.5x performance increase.

### DIFF
--- a/src/include/pipeline/hll.h
+++ b/src/include/pipeline/hll.h
@@ -29,12 +29,12 @@
 #define HLL_DENSE_CLEAN 'D'
 #define HLL_EXPLICIT_DIRTY 'e'
 #define HLL_EXPLICIT_CLEAN 'E'
-#define HLL_UNION 'u'
+#define HLL_UNPACKED 'u'
 
 #define HLL_IS_SPARSE(hll) ((hll)->encoding == HLL_SPARSE_DIRTY || (hll)->encoding == HLL_SPARSE_CLEAN)
 #define HLL_IS_EXPLICIT(hll) ((hll)->encoding == HLL_EXPLICIT_DIRTY || (hll)->encoding == HLL_EXPLICIT_CLEAN)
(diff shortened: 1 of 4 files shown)


src/include/pipeline/hll.h
```diff
@@ -29,10 +29,12 @@
 #define HLL_DENSE_CLEAN 'D'
 #define HLL_EXPLICIT_DIRTY 'e'
 #define HLL_EXPLICIT_CLEAN 'E'
+#define HLL_UNION 'u'
 
 #define HLL_IS_SPARSE(hll) ((hll)->encoding == HLL_SPARSE_DIRTY || (hll)->encoding == HLL_SPARSE_CLEAN)
 #define HLL_IS_EXPLICIT(hll) ((hll)->encoding == HLL_EXPLICIT_DIRTY || (hll)->encoding == HLL_EXPLICIT_CLEAN)
 #define HLL_IS_DENSE(hll) ((hll)->encoding == HLL_DENSE_DIRTY || (hll)->encoding == HLL_DENSE_CLEAN)
+#define HLL_IS_UNION(hll) ((hll)->encoding == HLL_UNION)
```
usmanm: I think calling this `UNION` is a little confusing since it's basically just a different encoding. We should maybe call it something like `UNPACKED` and then reuse functions like `HLLUnion` and just throw errors if `UNPACKED` is being merged with something other than `UNPACKED`.

PR #1789, opened 2017-04-01
Allow connecting the database in docker image from the other host

innossh:
We cannot connect to pipelinedb/pipelinedb:0.9.7 from the other host.

```
$ docker run -i -d -p 5432:5432 pipelinedb/pipelinedb:0.9.7
$ psql -h 127.0.0.1 -p 5432 -U pipeline
psql: server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
```

After this PR is merged, `listen_addresses` will be set "all", like this.

```
listen_addresses = '*'
#listen_addresses = 'localhost'         # what IP address(es) to listen on;
```


derekjn:
Thanks @innossh!

PR #1832, opened 2017-08-08
1702 activate

derekjn:


usmanm:
It might be worth starting to port functions over into a `.sql` file as well? Otherwise cleaning up the catalog files is going to become even more cumbersome?

A possible approach could be to create a new extension under `contrib/` called `pipeline` or something and bundle building/installing it with core, similar to how we had PostGIS back in the day.

derekjn:
> It might be worth starting to port functions over into a .sql file as well?

@usmanm definitely plan on doing that after pushing through some of this low-hanging fruit. ~200 functions is probably just as bad as ~200 + 2--it's going to be cumbersome no matter what :(

usmanm:
You're right. Cleaning up catalogs is going to be painful...

derekjn:
> You're right. Cleaning up catalogs is going to be painful...

It may be possible to automate some of it with a script that queries the DB, but either way it's going to take a couple of days of grinding.

PR #1844, opened 2017-08-11
1819 dest

derekjn:


derekjn:
@usmanm things are looking pretty clean. The physical tuple lookup path is fully standardized now, with no discernible perf impact. The addition of `CustomScan` and its associated planner hooks was a huge plus--we just missed it back in the day!

usmanm:
Dude, this is fucking sick. You're in beast mode!

Issue #1683, opened 2016-11-04
Replace custom syscache entries


derekjn:
We need our own lookup infrastructure that works with regular tables.

- [x] `PIPELINECOMBINEOID`
- [x] `PIPELINECOMBINETRANSFNOID`
- [x] `PIPELINEQUERYID`
- [x] `PIPELINEQUERYRELID`
- [x] `PIPELINEQUERYMATRELID`
- [x] `PIPELINEQUERYOID`
- [x] `PIPELINESTREAMRELID`
- [x] `PIPELINESTREAMOID`

derekjn:
We should make the interface nearly identical to the current syscache interface, so we can just rename function calls. e.g.: `SearchSysCache1` -> `SearchPipelineSysCache1`.

derekjn:
https://github.com/pipelinedb/pipelinedb/blob/master/src/backend/utils/cache/inval.c

PR #1846, opened 2017-08-12
#1683 Standardize syscache infrastructure

derekjn:


derekjn:
@usmanm so what I ended up doing here was basically just pulling all of our syscache entries into their own private module. The `PipelineSysCache*` and `SysCache*` functions then basically just do the exact same things, just operating on different private data structures. The catalog cache is actually pretty simple...

derekjn:
Also, the only real difference between the `PipelineSysCache` module and the original one is that the cache configuration is based on catalog and index name, rather than OID. This should be conducive to a fairly smooth migration once we create catalogs via a `.sql` files.

usmanm:
Interesting. So technically you can bolt on any table to the `CatCache` infrastructure? Is there any concern on the implementation details of `PipelineSysCache` diverging from core's `SysCache`?

derekjn:
> Interesting. So technically you can bolt on any table to the CatCache infrastructure?

It's based on the`systable_*` scan interface, which just needs to be passed an index so pretty much any table ([see these comments](https://github.com/pipelinedb/pipelinedb/blob/master/src/backend/access/index/genam.c#L295)).

>  Is there any concern on the implementation details of PipelineSysCache diverging from core's SysCache?

The actual implementation is shared. The only thing that's copied is the setup boilerplate (so that we can point ours to `PipelineSysCache` data structures), which would be pretty easy to change if ever necessary.

PR #1869, opened 2017-09-28
Restarts DB after a config file loaded via /docker-entrypoint.d/

jcrsilva:
is installed.

Some extensions have to be loaded via 'shared_preload_libraries',
like the kafka extension. If one requires views to be set up
when the container starts up via /docker-entrypoint-init.d/ with
the config being installed the same way, if the DB is not restarted
then these extensions are not available when the .sql files are run.


derekjn:
Thanks @jcrsilva!

Issue #1941, opened 2018-07-31
Allow the dockerfile's entrypoint to be called by child dockerfiles


joaosa:
I have a usecase where it'd be extremely useful to be able to reuse the `ENTRYPOINT` script from pipelinedb's dockerfile. 

This is currently possible with both [mysql](https://github.com/docker-library/mysql/blob/master/5.6/Dockerfile#L74) and [postgresql](https://github.com/docker-library/postgres/blob/master/9.5/Dockerfile#L147).
I propose a similar solution, i.e. simply adding `RUN ln -s usr/local/bin/docker-entrypoint.sh /entrypoint.sh # backwards compat` to pipelinedb's dockerfile.

PR #1942, opened 2018-07-31
Allow the dockerfile entrypoint to be reused by child images

joaosa:
Here's the associated issue: https://github.com/pipelinedb/pipelinedb/issues/1941

### DIFF
--- a/pkg/docker/Dockerfile
+++ b/pkg/docker/Dockerfile
@@ -155,6 +155,7 @@ RUN addgroup --system pipeline \
         pipeline
 
 COPY docker-entrypoint.sh /usr/local/bin/
+RUN ln -s usr/local/bin/docker-entrypoint.sh /entrypoint.sh # backwards compat
 RUN chmod +x /usr/local/bin/docker-entrypoint.sh
 
 ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]


derekjn:
@joaosa thank you so much! For reference purposes, do you mind briefly explaining what the issue is and why exactly this resolves it? Much appreciated! :pray:

joaosa:
You're welcome!

The main idea is to allow one to run any script before booting pipelinedb, while reusing pipelinedb's entrypoint and, thus, not having to boot it explicitly.

Here's an example:

dockerfile
```
FROM pipelinedb/pipelinedb:latest
ADD start.sh /
CMD ["/bin/sh", "/start.sh"]
```

start.sh
```
# do whatever you need to do
# e.g. fix permissions on a volume

# boot pipelinedb through its entrypoint
/entrypoint.sh pipelinedb
```

joaosa:
Turns out this is currently possible by calling `/usr/local/bin/docker-entrypoint.sh pipelinedb` instead.

It turns out `pipelinedb/pipelinedb:latest` isn't up to date and it seems to be a much older version. I'd say it's `0.9.6` (https://hub.docker.com/r/pipelinedb/pipelinedb/tags/).

joaosa:
Closing this as it's no longer relevant to me

Issue #1773, opened 2017-03-17
Building/Linking Ubuntu 16.10


jbylund:
Without making any modifications (from git head) on trying to build pipelinedb I get:
```
 ../../src/timezone/localtime.o ../../src/timezone/strftime.o ../../src/timezone/pgtz.o ../../src/port/libpgport_srv.a ../../src/common/libpgcommon_srv.a -lpthread /usr/lib/libzmq.a -lstdc++ -lrt -lcrypt -ldl -lm -o pipelinedb
gcc: error: /usr/lib/libzmq.a: No such file or directory
Makefile:60: recipe for target 'pipelinedb' failed
make[3]: *** [pipelinedb] Error 1
```
If I additionally create this link:
```
sudo ln --symbolic --force /usr/lib/x86_64-linux-gnu/libzmq.a /usr/lib/libzmq.a
```

I get a whole bunch of undefined symbol issues:
```
/usr/lib/libzmq.a(src_libzmq_la-ctx.o): In function `zmq::ctx_t::ctx_t()':
(.text+0x6ddd): undefined reference to `sodium_init'
/usr/lib/libzmq.a(src_libzmq_la-ctx.o): In function `zmq::ctx_t::~ctx_t()':
(.text+0x7491): undefined reference to `randombytes_close'
/usr/lib/libzmq.a(src_libzmq_la-zmq_utils.o): In function `zmq_curve_keypair':
(.text+0x36c): undefined reference to `crypto_box_keypair'
/usr/lib/libzmq.a(src_libzmq_la-zmq_utils.o): In function `zmq_curve_public':
(.text+0x3f4): undefined reference to `crypto_scalarmult_base'
/usr/lib/libzmq.a(src_libzmq_la-curve_client.o): In function `zmq::curve_client_t::encode(zmq::msg_t*)':
(.text+0x1c8): undefined reference to `crypto_box_afternm'
/usr/lib/libzmq.a(src_libzmq_la-curve_client.o): In function `zmq::curve_client_t::decode(zmq::msg_t*)':
```

Finally if I apply this diff:
```
diff --git a/src/Makefile.global.in b/src/Makefile.global.in
index 6404232..4b39317 100644
--- a/src/Makefile.global.in
+++ b/src/Makefile.global.in
@@ -600,7 +600,7 @@ TAS         = @TAS@
 #
 # Link to ZMQ
 
-LIBS := -lpthread /usr/lib/libzmq.a -lstdc++ $(LIBS)
[...]

PR #1989, opened 2018-11-18
ignore glibc specific code

docteurklein:


Makefile
```diff
@@ -9,7 +9,7 @@ REGRESS = $(EXTENSION)
 
 DATA = pipelinedb--1.0.0.sql
 EXTRA_CLEAN = src/test/regress/expected/$(REGRESS).out src/test/regress/sql/$(REGRESS).sql
-SHLIB_LINK += /usr/lib/libzmq.a -lstdc++
+SHLIB_LINK += -lzmq -lstdc++ -fPIC
```
docteurklein: found this on https://github.com/pipelinedb/pipelinedb/issues/1773#issuecomment-426924299
no idea what I'm doing

src/scheduler.c
```diff
@@ -6,7 +6,9 @@
  *
  *-------------------------------------------------------------------------
  */
+#ifdef GLIBC
```
docteurklein: again, not sure if that's even a good idea; found a similar fix in https://github.com/PowerDNS/pdns/pull/1958/files

src/scheduler.c
```diff
@@ -126,6 +128,7 @@ ContQueryDatabaseMetadataSize(void)
 void
 debug_segfault(SIGNAL_ARGS)
 {
+#ifdef GLIBC
```
docteurklein: maybe there is a way to find the "else" version for muslc ?

Makefile
docteurklein: I think this breaks the wanted static linking of libzmq in the extension, as I had to install  `zeromq-dev` in the runtime env.

docteurklein:
fxies #1692

docteurklein:
see #1578

dimitar-nikovski:
Excellent, many thanks for this solution and the edits, I wish there was an alpine-ready version out of the box.

binakot:
Thanks for this work, @docteurklein!
It was really helpful for my docker image https://github.com/binakot/PostgreSQL-TimescaleDB-PipelineDB. I know that just disable backtrace is not the solution, but at least it works. This was enough https://github.com/binakot/PostgreSQL-TimescaleDB-PipelineDB/blob/master/Dockerfile#L75

